PR #1, opened 2018-10-26
Add basic in-toto verifylib implementation

lukpueh:
See [`README.md`](https://github.com/lukpueh/in-toto-golang/blob/verifylib/README.md) for details about this PR.

Besides features that are available in the [Python reference implementation](https://github.com/in-toto/in-toto) but [not yet supported](https://github.com/lukpueh/in-toto-golang/tree/verifylib#not-yet-supported) in this implementation, we need:
* Better code documentation
* Better usage documentation
* Better user feedback
* More tests (Currently a full passing supply chain verification is tested, plus a few unit tests --> ~80% coverage)


in_toto/model.go
```diff
@@ -0,0 +1,188 @@
...
+  Environment map[string]interface{} `json:"environment"`
+}
+
+const LinkNameFormat = "%s.%.8s.link"
+const LinkNameFormatShort = "%s.link"
+
+type Inspection struct {
+  Type string `json:"_type"`
+  Run []string `json:"run"`
+  //TODO: Abstraction for Steps and Inspections?
```
aaaaalbert: Consistency: Other places read `// TODO:`.

in_toto/model.go
```diff
@@ -0,0 +1,188 @@
...
+    return err
+  }
+
+  // Write JSON bytes to the passed path with permissions (-rw-r--r--)
+  err = ioutil.WriteFile(path, jsonBytes, 0644)
+  if err != nil {
+    return err
+  }
+
+  return nil
```
aaaaalbert: Remove the `if` and implicit `else`, just `return err`.

in_toto/model.go
```diff
@@ -0,0 +1,188 @@
...
+  if sig == (Signature{}) {
+    return fmt.Errorf("No signature found for key '%s'", key.KeyId)
+  }
+
+  dataCanonical, err := mb.GetSignableRepresentation()
+  if err != nil {
+    return err
+  }
+
+  if err := VerifySignature(key, sig, dataCanonical); err != nil {
```
aaaaalbert: Is that an assignment inside a condition, like the old C `if foo = bar` mishap? If so, I suggest to split into separate lines (assuming there are no side effects).

Also, below there seems to be another superfluous check for non-`nil` before returning.

in_toto/rulelib.go
```diff
@@ -0,0 +1,99 @@
+package in_toto
+
+import (
+  "fmt"
+  "strings"
+  )
+
+var errorMsg string = "Wrong rule format, available formats are:\n\t" +
```
aaaaalbert: I don't like these trailing `\t`s to indent *the next line*.

in_toto/rulelib.go
```diff
@@ -0,0 +1,99 @@
...
+          "pattern": rule[1],
+        }, nil
+
+    case "match":
+      var srcPrefix string
+      var dstType string
+      var dstPrefix string
+      var dstName string
+
+      if ruleLen == 10 && ruleLower[2] == "in" &&
```
aaaaalbert: I'm not sure about hardcoded indices into a list. Consider using consts, like `IN_INDEX` and `WITH_INDEX`.

in_toto/rulelib.go
```diff
@@ -0,0 +1,99 @@
...
+      } else if ruleLen == 6 && ruleLower[2] == "with" &&
+          ruleLower[4] == "from" {
+        srcPrefix = ""
+        dstType = ruleLower[3]
+        dstPrefix = ""
+        dstName = rule[5]
+
+      } else {
+        return nil,
+            fmt.Errorf("%s Got:\n\t %s", errorMsg, rule)
```
aaaaalbert: Debugging feature request: Add source line number to error message, or otherwise disambiguate from an identical error message above.

aaaaalbert: ...and below, https://github.com/in-toto/in-toto-golang/pull/1/files#diff-c9f037ac0a1dba67bb7a5b123f411878R96

in_toto/rulelib_test.go
```diff
@@ -0,0 +1,80 @@
...
+        "dstType": "materials" , "dstName": "step-name"},
+    map[string]string{"type": "match", "pattern": "foo",
+        "srcPrefix": "", "dstPrefix": "dest-path",
+        "dstType": "products" , "dstName": "step-name"},
+    map[string]string{"type": "match", "pattern": "foo",
+        "srcPrefix": "", "dstPrefix": "",
+        "dstType": "materials" , "dstName": "step-name"},
+  }
+
+  for i := 0; i < len(rules); i++ {
```
aaaaalbert: Use "for statements with `range` clause" perhaps? https://golang.org/ref/spec#For_statements

in_toto/runlib.go
```diff
@@ -0,0 +1,136 @@
...
+  stdoutPipe, err := cmd.StdoutPipe()
+  if err != nil {
+    return nil, err
+  }
+
+  if err := cmd.Start(); err != nil {
+    return nil, err
+  }
+
+  // Todo duplicate stdout, stderr
```
aaaaalbert: Consistency --> `// TODO:`

in_toto/verifylib.go
```diff
@@ -0,0 +1,502 @@
...
+    // Dump inspection link to cwd using the short link name format
+    linkName := fmt.Sprintf(LinkNameFormatShort, inspection.Name)
+    linkMb.Dump(linkName)
+
+    inspectionMetadata[inspection.Name] = linkMb
+  }
+  return inspectionMetadata, nil
+}
+
+func Subtract(a []string, b []string) []string {
```
aaaaalbert: This is a set difference function in `O(n**2)`, right? If you expect large `a` and `b`, perhaps consider the linear-time version over here, https://siongui.github.io/2018/03/14/go-set-difference-of-two-arrays/ , or find a proper set library :-)

in_toto/verifylib.go
```diff
@@ -0,0 +1,502 @@
...
+        // Parse rule
+        ruleData, err := UnpackRule(rule)
+        if err != nil {
+          return err
+        }
+
+        // Process rules according to rule type
+        // TODO: Currently we only support "MATCH", "ALLOW" and "DISALLOW"
+        switch ruleData["type"] {
+          case "match":
```
aaaaalbert: Comment reads `MATCH`, this checks for lower-case `match`?

in_toto/verifylib.go
```diff
@@ -0,0 +1,502 @@
...
+  inspectionsI := make([]interface{}, len(layout.Inspect))
+  for i, v := range layout.Inspect {
+      inspectionsI[i] = v
+  }
+
+  if err := VerifyArtifacts(inspectionsI, inspectionMetadata); err != nil {
+    return err
+  }
+
+  return nil
```
aaaaalbert: Another `if`/ implicit `else` / actually returning `err`.

test/data/demo.layout.template
```diff
@@ -0,0 +1,136 @@
+{
+ "signatures": [
+  {
+   "keyid": "556caebdc0877eed53d419b60eddb1e57fa773e4e31d70698b588f3e9cc48b35",
+   "sig": "696c4f1e6d4de8b02cd94347f16536ffcd27c8b39b54d84884d205284e9594c6d88470c18f6c2b32259cbda9ebef3af933bc23031aeda25b681420df5b6da42c6fee132017e4d9156d3fef401c6a6d4a567848dc950a2d0874222b33ae4e66cd0aa3c0c9d33ed19e7fe54fe5b78404ed2a1ac2bb648dfe5c19ffe6837b4bb1ffb4b8eacd7ebb80caef1ad96b723636fdb84bccd076f49d4fefce2c09dbc1b297c1d38c08a48ed1c5231c711722d744f33a7bc06b5291270f8e3e8cf3c2ee28aba53a6cd62e1171ba8c77b43d6fd85dada544b606e52c7bab840e0ff2793f2fdc0fa86c018c8e37dc0cd0ea23a4678c1a7689b58f710bbcd81f3a4f1b04b130978aaa77727f8afdd385b5a6dfc36e6ea638bf74c68acdfd00e72dd6614b20d57ab99e36ea171657eb2694cbecb4665eb8006fbe8e03d9c5e090d0c19c6a93db484a59c0ba9c2bb2bcb0e5c69c4732de3365859f52ac75c49840601155c4a1bebadf2759db9f308bdb0b5d628177ee7cea6953a3f31049b8f52b3c5007c5fe1cbb"
+  }
+ ],
+ "signed": {
+  "_type": "layout",
+  "expires": "2018-11-18T16:06:36Z",
```
aaaaalbert: Happy testing in three weeks' time :-)

aaaaalbert: On a more serious note, can't / shoudn't the keys, tarballs etc. be created on-the-fly for testing?

lukpueh: Good eyes! But did you not see the note in commit message for f0c23f1e37a92ae4538217bbc02c63c425614d25?! (just kidding) You're right, these things (layout expiration date and signature) should be added on the fly, as we do in the reference implementation tests. The only reason why I hardcoded the date and pre-signed the layout for now is because I haven't implemented signing yet (not required for in-toto verification). I set myself a three weeks time window to fix this. :)

I also meant to add a comment, but there's no commenting in JSON. I could add a todo note in `test_verifylib.go`.

in_toto/rulelib.go
lukpueh: Would you prefer the `\t` on the next line, or should I remove it altogether?

in_toto/rulelib.go
lukpueh: Valid point. On a more general note, I am still looking for a good way to propagate error messages, e.g. stack traces for debugging (with line numbers and everything) and nice and precise messages for user feedback. Something like...

```go
func a() error {
  return fmt.Errorf("THE ERROR")
}

func b() error {
  if err := a(); err != nil {
    return fmt.Errorf("calling a: %s", err)
  }
  return nil
}

if err := b(); err != nil  {
  fmt.Printf("calling b: %s", err)
}
```

Any suggestions? Here's what I know about error handling in go:
https://blog.golang.org/error-handling-and-go
https://opencredo.com/why-i-dont-like-error-handling-in-go/

in_toto/model.go
lukpueh: The reason why I wrote it that way, is because I thought about (later) enriching the error message with some context information. See https://github.com/in-toto/in-toto-golang/pull/1#discussion_r230003010 for more thoughts on this. I will summarize your comments and my thoughts in a GitHub issue, if that's okay with you.

in_toto/model.go
lukpueh: Re superfluous check for non-nil
See https://github.com/in-toto/in-toto-golang/pull/1#discussion_r230004116

in_toto/verifylib.go
lukpueh: See https://github.com/in-toto/in-toto-golang/pull/1#discussion_r230004116

in_toto/rulelib.go
lukpueh: Not sure if consts make sense, because the token's may have different positions in the list.
I hope these comments help: 486e53d027a54f2c94fd5eddc8553e2b8444cf72

in_toto/rulelib_test.go
lukpueh: Oh yes, makes perfect sense. Fixed in 20f9e1e9f6dbe07fa0aeca68e4d320d3da5b7992

in_toto/verifylib.go
lukpueh: Oh yes, I was considering using the more performant map *hack*, but opted for the other one for reasons of readability, and because I thought `O(n**2)` is *okay* for now. But you're right. I will at least add a comment.

test/data/demo.layout.template
aaaaalbert: How could I miss that :-)

Let's add a todo note, or file an issue.

lukpueh: I added one in the verifylib test da9a05e

in_toto/model.go
aaaaalbert: Understood. (I wonder if that's an instance of YAGNI, and/or how you will grep for a puzzling multi-line statement when later wanting to replace it, rather than a plan `return err`.)

in_toto/model.go
aaaaalbert: This is less often *used* in C I fear, it's mostly an *error* on behalf of the programmer :-)

I'm fine if this is considered sensible Go style.

in_toto/rulelib.go
aaaaalbert: On the next line.

in_toto/rulelib.go
aaaaalbert: Yeah, I see, I too felt my suggestion wasn't very pretty.

A link to the spec perhaps makes sense, so that when this code gets updated, it's clear what version etc. this implementation supports.

aaaaalbert: (My original comment was coming from the idea of future-proofing the field numbers.)

in_toto/rulelib.go
aaaaalbert: Let's @-mention the experts then.

in_toto/verifylib.go
aaaaalbert: Saw your comment in 0d6b3c065c931996ba62eac43fe58f5fb7e50876 and liked it!

in_toto/verifylib.go
aaaaalbert: OK, rules are case-insensitive (db1496bb7ab75a4e72a6191a270c835e67b16ce9), but are the contents of `ruleData`'s entries sure to be lower-cased?

in_toto/verifylib.go
```diff
@@ -0,0 +1,517 @@
...
+    for _, verificationData := range verificationDataList {
+
+      rules := verificationData["rules"].([][]string)
+      artifacts := verificationData["artifacts"].(map[string]interface{})
+
+      // Create a queue of artifact names (paths)
+      // Each rule only operates on artifacts in that queue
+      // If a rule consumes an artifact (can be applied successfully) it is
+      // removed from the queue.
+      // By applying a DISALLOW rule eventually, verification may return an
```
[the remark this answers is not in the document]
lukpueh: :P 

`DISALLOW` is the general (platonic) idea of that rule. Such a rule might be summarized in a `ruleData` map where `ruleData["type"]` is `"disallow"`.

in_toto/verifylib.go
lukpueh: I  addressed it in ca976fe. Not sure why the conversation is not marked as outdated.

aaaaalbert: ca976fe addresses the case of words in the comment string; what about the contents of the variable at runtime? Where do they come from? Are the forced to lower case somewhere?

in_toto/verifylib.go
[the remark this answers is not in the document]
aaaaalbert: :+1: for Plato.

in_toto/verifylib.go
aaaaalbert: Like https://github.com/in-toto/in-toto-golang/pull/1/files#diff-5c3ce1a070ae336476f0f76a5decd5c5R106 , does this upper-case `MATCH` ever end up in `ruleData`?

lukpueh: `ruleData` is assigned by calling `rulelib.UnpackRule` above:
 https://github.com/in-toto/in-toto-golang/blob/ca976fee048252062a307af77fe69ac323056425/in_toto/verifylib.go#L138

That function normalizes the case of rule tokens as they appear in the layout.
https://github.com/in-toto/in-toto-golang/blob/ca976fee048252062a307af77fe69ac323056425/in_toto/rulelib.go#L23-L29

aaaaalbert: Thank you very much, I *finally* get it :-)

in_toto/verifylib.go
```diff
@@ -0,0 +1,517 @@
...
+                  strings.ToUpper(ruleData["type"]))
+        }
+      }
+    }
+  }
+  return nil
+}
+
+
+func ReduceStepsMetadata(layout Layout,
```
SantiagoTorres: What does Reducing steps metadata actually entail?

in_toto/verifylib.go
```diff
@@ -0,0 +1,517 @@
...
+              fmt.Sprintf(LinkNameFormat, step.Name, referenceKeyID),
+              fmt.Sprintf(LinkNameFormat, step.Name, keyID))
+        }
+      }
+      // We haven't errored out, so we can reduce
+      stepsMetadataReduced[step.Name] = referenceLinkMb
+    }
+  }
+  return stepsMetadataReduced, nil
+}
```
SantiagoTorres: I somehow think this is a threshold signature check. Should we name it like so?

in_toto/verifylib.go
```diff
@@ -0,0 +1,517 @@
...
+          "', no link metadata found.")
+    }
+
+    for signerKeyID, linkMb := range linksPerStep {
+      expectedCommandS := strings.Join(step.ExpectedCommand, " ")
+      executedCommandS := strings.Join(linkMb.Signed.(Link).Command, " ")
+
+      if expectedCommandS != executedCommandS {
+        linkName := fmt.Sprintf(LinkNameFormat, step.Name, signerKeyID)
+        fmt.Printf("WARNING: Expected command for step '%s' (%s) and command" +
```
SantiagoTorres: I wonder if we want to print to stderr instead.

in_toto/verifylib.go
```diff
@@ -0,0 +1,517 @@
...
+  // rules may also refer to artifacts reported by step links
+  for k, v := range stepsMetadataReduced {
+    inspectionMetadata[k] = v
+  }
+
+  // Convert []Inspections to []interace{} (see `stepsI` above)
+  inspectionsI := make([]interface{}, len(layout.Inspect))
+  for i, v := range layout.Inspect {
+      inspectionsI[i] = v
+  }
```
SantiagoTorres: I wonder if all of this stupid lang-related type juggling can be moved to helper functions to simplify things.

in_toto/rulelib.go
```diff
@@ -0,0 +1,107 @@
...
+func UnpackRule(rule []string) (map[string]string, error) {
+  // Cache rule len
+  ruleLen := len(rule)
+
+  // Create all lower rule copy to case insensitively parse out tokens whose
+  // position we don't know yet
+  // We keep the original rule to retain the non-token elements' case
+  ruleLower := make([]string, ruleLen)
+  for i, val := range rule {
+    ruleLower[i] = strings.ToLower(val)
```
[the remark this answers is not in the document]
aaaaalbert: +:100:

in_toto/model.go
aaaaalbert: ("Golang-uistic"?)

in_toto/rulelib.go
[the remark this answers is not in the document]
lukpueh: Agreed, that's why I only use `ruleLower` to parse out the tokens (`<RULE TYPE>`, `IN`, `WITH`, `<ARTIFACT TYPE>`,`FROM`). Non-token elements (`<pattern>`, `<path prefix>`, `<step name>`) are taken from the original non-normalized `rule` slice.

Should I update the comment in L23-L25?

in_toto/keylib.go
```diff
@@ -0,0 +1,137 @@
...
+
+  // TODO: There could be more key data in _, which we silently ignore here.
+  // Should we handle it / fail / say something about it?
+  data, _ := pem.Decode([]byte(keyBytes))
+  if data == nil {
+    return fmt.Errorf("No valid public rsa key found at '%s'", path)
+  }
+
+  // Try if this is a indeed a public key, just to see if the key is valid
+  _, err = x509.ParsePKIXPublicKey(data.Bytes)
```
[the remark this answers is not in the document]
lukpueh: It does not, `x509.ParsePKIXPublicKey` returns an `interface` that may be a struct of type `rsa.PublicKey`, `dsa.PublicKey`, or `ecdsa.PublicKey`, but we need to store the key in the original pem format, extracted by the next ~15 lines.

I only call `pem.Decode` and `x509.ParsePKIXPublicKey` to fail early if the data can't be parsed. Does that make sense?

I could reconstruct the original pem, by using the data returned by one of those two functions, but it would not reduce code.

in_toto/keylib.go
```diff
@@ -0,0 +1,137 @@
...
+  if keyStart == -1 || keyEnd == -1 {
+    return fmt.Errorf("No valid public rsa key found at '%s'", path)
+  }
+  keyBytesStripped := keyBytes[keyStart:keyEnd]
+
+  // Declare values for key
+  // TODO: Do not hardcode here, but define defaults elsewhere and add support
+  // for parametrization
+  keyType := "rsa"
+  scheme := "rsassa-pss-sha256"
```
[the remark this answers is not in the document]
lukpueh: The [`Block`](https://golang.org/pkg/encoding/pem/#Block) returned by `pem.Decode` has a `Type` field, which is the string taken from the pem preamble (e.g. `"RSA PRIVATE KEY"`).

I could (and probably should) check the returned data of `x509.ParsePKIXPublicKey` to see if it is rsa, dsa or ecdsa.

in_toto/verifylib.go
lukpueh: Yes good idea. Is it okay, if I create a ticket for this?

in_toto/verifylib.go
lukpueh: Yes! I wanted to rework user output and logging anyway. Haven't figured out yet how to best do this in go. :) Is it okay, if I create a ticket for this?

in_toto/verifylib.go
lukpueh: It takes multiple links corresponding to a given step of the layout and merges them into a single link to make further processing (artifact rules) easier. If the links have different materials or different products it fails.

IIRC you, @RohanBhirangi and I tried hard to find a proper name for this function, when Rohan added it to the reference implementation, and this was the best compromise we could come up with.

I am still open for alternative names.

in_toto/verifylib.go
lukpueh: Rather not, there is a separate `VerifyLinkSignatureThesholds` function that gets called earlier. We could merge `VerifyLinkSignatureThesholds` and `ReduceStepsMetadata`, but then we can't verify command alignment for all links anymore, which we do after verifying signature thresholds, but before merging all links for a step.

in_toto/runlib.go
```diff
@@ -0,0 +1,138 @@
...
+  // See https://stackoverflow.com/questions/10385551/get-exit-code-go
+  if err != nil {
+    if exiterr, ok := err.(*exec.ExitError); ok {
+      // The program has exited with an exit code != 0
+      // This works on both Unix and Windows. Although package
+      // syscall is generally platform dependent, WaitStatus is
+      // defined for both Unix and Windows and in both cases has
+      // an ExitStatus() method with the same signature.
+      if status, ok := exiterr.Sys().(syscall.WaitStatus); ok {
+        retVal = status.ExitStatus()
```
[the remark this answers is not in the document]
lukpueh: yeah, this is weird.

in_toto/rulelib_test.go
```diff
@@ -0,0 +1,80 @@
...
+              returnedRuleMap[key])
+      }
+    }
+  }
+}
+
+func TestUnpackInvalidRules(t *testing.T) {
+  rules := [][]string {
+    []string{"CREATE", "foo", "too-long"},
+    []string{"SUBVERT", "foo"},
```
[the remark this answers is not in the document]
lukpueh: let's add it to the specs then!

in_toto/keylib.go
```diff
@@ -0,0 +1,137 @@
...
+    }
+  k.Scheme = scheme
+  k.KeyIdHashAlgorithms = keyIdHashAlgorithms
+  k.KeyId = fmt.Sprintf("%x", keyHashed)
+
+  return nil
+}
+
+
+func VerifySignature(key Key, sig Signature, data []byte) error {
```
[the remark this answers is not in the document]
lukpueh: Yes, is it okay if we make a separate cleanup PR for this?

in_toto/model.go
lukpueh: What would be the `value`? `Dump` only has a write side-effect but doesn't return anything (other than an error).

in_toto/model.go
lukpueh: Same as answer to your comment above. What should the first value in the return tuple contain?

`VerifySignature` here is similar to the corresponding function in the Python reference implementation that returns nothing (on success) or raises an exception.

Re simplification:
Two against one, I surrender. :)
Is it okay if I replace expression-in-condition constructs everywhere in a separate PR?

aaaaalbert: 1, it's only 1.5 against 1 because this seems to be a Go idiom.

2, I'm fine with a separate PR.

in_toto/rulelib.go
lukpueh: It looks pretty much the same as in the reference implementation. While that is no argument for it, it is an indication that we haven't found (and needed) a better solution yet. Do you want to create a ticket, with an implementation suggestion? I'm not sure if I understand your suggestion here.

in_toto/keylib.go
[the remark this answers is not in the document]
lukpueh: Added an rsa check in 61a1f3cf7da0e03e3254f75914ec8cad9fab83a2

lukpueh:
Thanks for your awesome reviews @SantiagoTorres and @aaaaalbert! Please let me know if creating issues for some of the comments you left is okay, or if you'd rather have me fix the things right away.

aaaaalbert:
My pleasure :-)

I suggest you create issues for all the other things so that this PR isn't left open for too much longer.

SantiagoTorres:
I'm fine with creating issues :)

in_toto/rulelib.go
lukpueh: Will be dealt with in https://github.com/in-toto/in-toto-golang/issues/4

in_toto/verifylib.go
lukpueh: See https://github.com/in-toto/in-toto-golang/issues/4

in_toto/rulelib.go
lukpueh: See https://github.com/in-toto/in-toto-golang/issues/5

in_toto/verifylib.go
lukpueh: See https://github.com/in-toto/in-toto-golang/issues/6

Issue #7, opened 2018-11-12
Refactor VerifySignature


lukpueh:
@SantiagoTorres suggests to factor out some functionality of `keylib`'s `VerifySignature` into helpers for the Key struct, see https://github.com/in-toto/in-toto-golang/pull/1#discussion_r232053885.

lukpueh:
Fixed with #12

Issue #6, opened 2018-11-12
Create helper functions for *lang-related type juggling* 


lukpueh:
Move the following two type conversion loops to helper functions:
https://github.com/in-toto/in-toto-golang/blob/ca976fee048252062a307af77fe69ac323056425/in_toto/verifylib.go#L480-L486

https://github.com/in-toto/in-toto-golang/blob/ca976fee048252062a307af77fe69ac323056425/in_toto/verifylib.go#L503-L507

Addresses @SantiagoTorres's https://github.com/in-toto/in-toto-golang/pull/1#discussion_r232062368.

lukpueh:
Fixed with #12

PR #12, opened 2018-11-14
Clean up InTotoVerify and rsa public key loading

lukpueh:
This PR provides a few small code refactors for clean up:
- Summarize common functionality used in keylib VerifySignature and LoadPublicKey #7 
- Move type juggeling loops from InTotoVerify to helper methods #6 
- Implement go-thonic field inheritance for common fields of steps/inspections

### DIFF
--- a/in_toto/keylib.go
+++ b/in_toto/keylib.go
@@ -14,6 +14,29 @@ import (
   "reflect"
 )
 
+func ParseRSAPublicKeyFromPEM(pemBytes []byte) (*rsa.PublicKey, error) {
+  // TODO: There could be more key data in _, which we silently ignore here.
+  // Should we handle it / fail / say something about it?
+  data, _ := pem.Decode([]byte(pemBytes))
+  if data == nil {
+    return nil, fmt.Errorf("Could not find a public key PEM block")
+  }
+
+  pub, err := x509.ParsePKIXPublicKey(data.Bytes)
+  if err != nil {
+    return nil, err
+  }
+
+  //ParsePKIXPublicKey might return an rsa, dsa, or ecdsa public key
+  rsaPub, isRsa := pub.(*rsa.PublicKey)
+  if !isRsa {
+    return nil, fmt.Errorf("We currently only support rsa keys: got '%s'",
+        reflect.TypeOf(pub))
+  }
+
+  return rsaPub, nil
+}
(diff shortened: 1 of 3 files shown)


in_toto/keylib.go
```diff
@@ -14,6 +14,29 @@ import (
   "reflect"
 )
 
+func ParseRSAPublicKeyFromPEM(pemBytes []byte) (*rsa.PublicKey, error) {
+  // TODO: There could be more key data in _, which we silently ignore here.
+  // Should we handle it / fail / say something about it?
```
SantiagoTorres: If we were to merge this, could we ticketize this issue?

in_toto/model.go
```diff
@@ -68,6 +68,27 @@ type Layout struct {
   Readme string `json:"readme"`
 }
 
+// Go does not allow to pass `[]T` (slice with certain type) to a function
+// the accepts `[]interface{}` (slice with generic type)
```
SantiagoTorres: s/the/that/

in_toto/model.go
```diff
@@ -68,6 +68,27 @@ type Layout struct {
...
+  }
+  return stepsI
+}
+func (l *Layout) InspectAsInterfaceSlice() []interface{} {
+  inspectionsI := make([]interface{}, len(l.Inspect))
+  for i, v := range l.Inspect {
+    inspectionsI[i] = v
+  }
+  return inspectionsI
+}
```
SantiagoTorres: Aside: is this just another instance of lolnogenerics?

in_toto/keylib.go
lukpueh: Yes. Done. #14.

in_toto/model.go
lukpueh: :)

in_toto/model.go
lukpueh: a685b474ff64e059aabbc1feb125c899305db28c

lukpueh:
@SantiagoTorres, thanks for your review! I addressed you comments, except for the one about lolnogenerics, I fear I can't fix that one. :P  I also had to quickly fix the test layout expiration (#15) and rebase.

SantiagoTorres:
Thanks @lukpueh !

PR #19, opened 2019-02-07
Add sublayout verification capabilities

adityasaky:
This PR introduces capabilities for sublayout verification. There are still a couple of things left to do, but I went ahead and opened this to start any relevant discussions.

Things left to do:
1. Add test for  multi level sublayouts
~~2. Add some comments~~
~~3. Run the files through gofmt~~
~~4. Line wrapping~~


adityasaky:
Note: Also looking into why builds failed.

in_toto/verifylib_test.go
```diff
@@ -62,16 +69,120 @@ func TestInTotoVerifyPass(t *testing.T) {
...
+	var err error
+	if summaryLink, err = GetSummaryLink(demoLayout.Signed.(Layout),
+		demoLink); err != nil {
+		t.Error(err)
+	}
+	if summaryLink.Signed.(Link).Type != codeLink.Signed.(Link).Type {
+		t.Errorf("Summary Link isn't of type Link")
+	}
+	if !reflect.DeepEqual(summaryLink.Signed.(Link).Name,
+		codeLink.Signed.(Link).Name) {
```
lukpueh: No need to use `DeepEqual` for string comparison.

in_toto/verifylib_test.go
```diff
@@ -62,16 +69,120 @@ func TestInTotoVerifyPass(t *testing.T) {
...
+		t.Errorf("Summary Link by-products don't match. Expected '%s', "+
+			"returned '%s", packageLink.Signed.(Link).ByProducts,
+			summaryLink.Signed.(Link).ByProducts)
+	}
+	if !reflect.DeepEqual(summaryLink.Signed.(Link).ByProducts["return-value"],
+		packageLink.Signed.(Link).ByProducts["return-value"]) {
+		t.Errorf("Summary Link return value doesn't match. Expected '%s', "+
+			"returned '%s", packageLink.Signed.(Link).ByProducts["return-value"],
+			summaryLink.Signed.(Link).ByProducts["return-value"])
+	}
```
lukpueh: No need to compare `"return-value"` in `ByProducts`, if you've already compared `ByProducts`.

in_toto/verifylib_test.go
```diff
@@ -62,16 +69,120 @@ func TestInTotoVerifyPass(t *testing.T) {
...
+	packagePath := path.Join(sublayoutDirectory, "package.2f89b927.link")
+	if err := os.Link("package.2f89b927.link", packagePath); err != nil {
+		t.Errorf("Unable to link package metadata.")
+	}
+
+	var superLayoutMb Metablock
+	if err := superLayoutMb.Load("super.layout"); err != nil {
+		t.Errorf("Unable to load super layout")
 	}
+
```
lukpueh: Can you also call `InTotoVerify` on the super layout to test the nesting?

lukpueh: And can you also create a test case, where super layout verification fails, because sublayout verification fails?

in_toto/verifylib_test.go
```diff
@@ -62,16 +69,120 @@ func TestInTotoVerifyPass(t *testing.T) {
...
+	stepsMetadataVerified := make(map[string]map[string]Metablock)
+	if stepsMetadataVerified, err = VerifyLinkSignatureThesholds(
+		superLayoutMb.Signed.(Layout), stepsMetadata); err != nil {
+		t.Errorf("Unable to verify link threshold values.")
+	}
+
+	if _, err := VerifySublayouts(superLayoutMb.Signed.(Layout),
+		stepsMetadataVerified, "."); err != nil {
+		t.Errorf("Unable to verify sublayouts.")
+	}
```
lukpueh: It would also be nice to take a look at the result to see if it returns the right thing.

in_toto/verifylib.go
```diff
@@ -557,67 +627,64 @@ is returned.
...
 		stepsMetadata)
 	if err != nil {
-		return err
+		return temp, err
 	}
 
-	// TODO: Verify sublayouts
+	// Verify and resolve sublayouts
+	stepsSublayoutVerified, err := VerifySublayouts(layout,
+		stepsMetadataVerified, linkDir)
```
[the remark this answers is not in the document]
adityasaky: I added a check for err != nil for now, I'm leaving this unresolved wrt the second comment.

Apologies for missing the fairly basic check.

in_toto/verifylib_test.go
adityasaky: That would require throwing in a bunch of sublayout files into the repository. I was thinking of working on runlib next and at the very least adding signing capabilities so we can generate what we need for the tests. Maybe this can be done then?

lukpueh: That of course would be really nice!

adityasaky: Okay, I'll place it on the back burner for now then. I think working on runlib will give me a cleaner picture of it too, which is very useful. I suppose we should hold off on merging this till then?

in_toto/verifylib.go
```diff
@@ -557,67 +627,64 @@ is returned.
...
+func InTotoVerify(layoutMb Metablock, layoutKeys map[string]Key,
+	linkDir string) (Metablock, error) {
 
-	var layoutMb Metablock
-
-	// Load layout
-	if err := layoutMb.Load(layoutPath); err != nil {
-		return err
-	}
+	var temp Metablock
```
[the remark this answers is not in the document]
adityasaky: Go doesn't allow us to return `nil` in these situations. I've restructured it slightly to use `summaryLink` itself, albeit it'll be empty.

in_toto/verifylib_test.go
adityasaky: What would such a check look like? Should we verify the data in the `map`?

in_toto/verifylib.go
```diff
@@ -533,19 +533,93 @@ func VerifyLayoutSignatures(layoutMb Metablock,
...
+		/* Using the name of the first link for the summary link can be misleading
+		as it's likely not descriptive of all the steps in the sublayout. The first step
+		of the sublayout may also be rather trivial, such as extracting an archive, and
+		it could also be non-unique. The superlayout and the sublayout may share a similar
+		step which can be confusing.
+
+		It's possible to chain together all the step names into a string as follows:
+		"<step_name>, <step_name>...",
+		but this can get very long if we have nested sublayouts.
+		*/
```
lukpueh: Thanks for bringing this to our attention. It is something that also needs fixing in the reference implementation. I created a ticket at https://github.com/in-toto/in-toto/issues/254. Please consider applying my proposed fix for this issue here.

adityasaky:
Latest commits potentially fix in-toto/in-toto#254, if accepted, I'll push a similar fix to the reference implementation.

in_toto/examples_test.go
```diff
@@ -5,9 +5,9 @@ import (
 	"os"
 )
```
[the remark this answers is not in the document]
adityasaky: They need to be in the `test/data` directory. `go test` will copy them into the `pwd` which is created for that particular test instance.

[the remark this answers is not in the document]
lukpueh: Yes, that is true when you run the code with `go test`. But `example_*.go` files are also rendered as sample snippets in the auto-generated go docs (try to run e.g. `godoc -http :6060`, and browse [your localhost](http://localhost:6060/pkg/github.com/in-toto/in-toto-golang/in_toto/#example_InTotoVerify)).
Someone reading that  stand-alone sample code might need some additional context, hence the comment.

lukpueh:
Great work, @adityasaky! Let's merge and add more tests with #22

Issue #21, opened 2019-02-11
Add metadata schema validation


lukpueh:
**Description of issue or feature request**:
`model.go` defines structs that may be used to hold in-toto metadata. The metadata schema is only loosely prescribed by the used datatypes and by the specified json tags.

We need a generic schema validation facility that allows to verify if a given instance of a struct is conformant with the [in-toto metadata specification](https://github.com/in-toto/docs/blob/v0.9/in-toto-spec.md#4-document-formats). 

For inspiration take a look at the reference implementation, where we use a [`ValidationMixin`](https://github.com/in-toto/in-toto/blob/v0.2.3/in_toto/models/common.py#L27-L54) for top-level metadata models, plus [`securesystemslib.schema`](https://github.com/secure-systems-lab/securesystemslib/blob/v0.11.2/securesystemslib/schema.py)-checkers to validate mostly (but not exclusively) crypto-related data structures, see as defined in ([`securesystemslib.formats`](https://github.com/secure-systems-lab/securesystemslib/blob/v0.11.2/securesystemslib/formats.py), [`in_toto.formats`](https://github.com/in-toto/in-toto/blob/v0.2.3/in_toto/formats.py) and [`in_toto.gpg.formats`](https://github.com/in-toto/in-toto/blob/v0.2.3/in_toto/gpg/formats.py)).

**Current behavior**:
No methods to validate the data model.

**Expected behavior**:
Provide methods to validate the data model and use where appropriate (e.g. to check function parameters).

adityasaky:
I'm working on this.

adityasaky:
I was looking at the validations being carried out in the reference implementation, specifically in `Layout` and `Link`.

- In both, there's a check for the value of `type` is what it should be - this is a valid check.
- In `Layout`, the expiry date is validated - both the format and if it's a an actual valid date. Again, this is a valid check.
- In `Layout`, readme is checked to ensure it's a string - this is redundant in golang as far as I can tell.
- In `Layout`, the keys are checked mostly for the format of the keyid and the key itself. 
- In `Layout`, the steps and inspections are checked. The checks are for type and uniqueness - again, checking for the type feels redundant in golang. Checking uniqueness is definitely important.
- In `Link`, the materials, products, and byproducts are checked for their type, and are independently validated. The type check again feels redundant, however, I'm going to look at the validate calls for steps and inspections and see what's happening there next.

[...]

PR #25, opened 2019-03-06
Add validation functions

adityasaky:
**Fixes issue #**: #21 

This introduces validation functions for metadata schema, similar to those present in the reference implementation. As of opening this PR, it's still missing checks for disallowed fields and missing compulsory fields. See in-toto/in-toto#186.


adityasaky:
I've opened this up because I wanted to discuss the best way to implement a couple of things:

1. As there aren't constructors per se, when must `validate()` be called? Should we expect a manual call to `validate()` every time a new instance is created?

2. While I initially threw around the idea of using [DisallowUnknownFields](https://golang.org/pkg/encoding/json/#Decoder.DisallowUnknownFields), it would mean moving away from `Unmarshal` which I don't like because `Decoder` typically expects a stream of data, which isn't what we have here. Our other option is to use a map to maintain what the spec is, and what it expects, and match metadata with it. While an array would probably suffice, using a map allows us to support optional and compulsory fields.

Point 2 is summarized partly from a conversation @SantiagoTorres and I had yesterday where we discussed maps vs arrays for this context.

in_toto/model.go
```diff
@@ -150,6 +190,67 @@ func (l *Layout) InspectAsInterfaceSlice() []interface{} {
 	return inspectionsI
 }
 
+func (l *Layout) validate() {
```
SantiagoTorres: This is not a stab at lolnogenerics, but can we generalize this function?

SantiagoTorres: Also, it feels like these functions should probably be throwing an error

adityasaky: Yep, I need to update them to return errors.

lukpueh: +1 generalization
+1 returning errors

in_toto/model.go
```diff
@@ -150,6 +190,67 @@ func (l *Layout) InspectAsInterfaceSlice() []interface{} {
...
+			fmt.Println("Unable to check if key ID has valid format")
+		}
+		if !keyIdFormatCheck {
+			fmt.Println("Key ID has invalid format")
+		}
+
+		if key.KeyId != keyId {
+			fmt.Println("Invalid key found")
+		}
+		// SSL Sets this to be optional. Can it ever be non empty?
```
lukpueh: Yes and no. In a layout (i.e. here) the private portion of the key must be empty, but in general a `Key` can have public and private portions.

in_toto/model.go
```diff
@@ -150,6 +190,67 @@ func (l *Layout) InspectAsInterfaceSlice() []interface{} {
...
+
+func (l *Layout) validateExpires() {
+	if _, err := time.Parse("2006-01-02T15:04:05Z", l.Expires); err != nil {
+		fmt.Println("Expiry time parsed incorrectly - date either invalid or" +
+			" of incorrect format")
+	}
+}
+
+func (l *Layout) validateKeys() {
+	for keyId, key := range l.Keys {
```
lukpueh: If we had a `Key` validator, we could use it here and elsewhere. What do you think? 
So instead of below code you just do

```go
for keyId, key := range l.Keys {
  key.validate()
...
```

in_toto/model.go
```diff
@@ -150,6 +190,67 @@ func (l *Layout) InspectAsInterfaceSlice() []interface{} {
...
+}
+
+func (l *Layout) validateType() {
+	if l.Type != "layout" {
+		fmt.Println("Invalid Type value for layout: should be 'layout'")
+	}
+}
+
+func (l *Layout) validateExpires() {
+	if _, err := time.Parse("2006-01-02T15:04:05Z", l.Expires); err != nil {
```
lukpueh: We should have a constant for our own *profile* of ISO 8601, which is similar to RFC 3339 but without trailing offset (as you correctly hard-coded here). In `securesystemslib` it is called [`ISO8601_DATETIME_SCHEMA`](https://github.com/secure-systems-lab/securesystemslib/blob/v0.11.2/securesystemslib/formats.py#L93). The same one could then be used in `verifylib.VerifyLayoutExpiration`.

in_toto/model.go
```diff
@@ -112,6 +127,31 @@ type Step struct {
...
+
+func (s *Step) validateType() {
+	if s.Type != "step" {
+		fmt.Println("Invalid Type value for step: should be 'step'")
+	}
+}
+
+func (s *Step) validatePubkeys() {
+	for _, keyId := range s.PubKeys {
+		keyIdFormatCheck, err := regexp.MatchString("[a-fA-F0-9]+", keyId)
```
lukpueh: I think this should be a separate key id validator, to be re-used with e.g. a `Key` struct validator.

in_toto/model.go
adityasaky: Good call.

in_toto/model.go
adityasaky: Okay, I'll do that. It'll just be set to the same value I've used, as that's what `time.Parse` expects.

adityasaky:
@lukpueh

I'm looking at the validator library and some thoughts:
- It will take care of the when to call validate problem
- It will make a few of the functions redundant, though I think some will need to remain

However, given that we have a handful of validation functions right now, I'm not sure it's necessary any time soon. I guess it also comes down to the roadmap and if we can expect a significant increase in validations.

in_toto/model.go
adityasaky: Do you have an idea for how to distinguish between public and private keys? `key.validate` must check that the private part is empty for a public key, and I was wondering if that should just be a separate check? If it is, it doesn't really leave enough validation processes to warrant being a part of the `Key` struct I think.

lukpueh:
Agreed! Let's roll our own validation.

in_toto/model.go
lukpueh: Not sure, I understand what you mean by your last sentence. Here are some thoughts:

In the reference implementation we use an OOP-approach for the top level metadata, e.g. if you call `layout.validate()` on a `Layout` object, it knows how to validate itself.

For lower level metadata we use securesystemslib's approach, which although implemented with OOP is actually functional in regard to the validated data, i.e. if you want to validate a key, it does not know what kind of key it is or that it is a key at all, so you must specify as what you want to validate it, e.g. `PUBLIC_KEY_SCHEMA.matches(something)`.

AFAICS go lends itself towards the second approach, because it has very stripped-down OOP concepts (e.g. no inheritance). So to get back to your original question, I'd probably go for multiple key validation functions, along the lines `validateAnyKey`, `validatePubKey`, `validatePrivKey`.

It might also be worth re-considering the general use of receivers, e.g. `key.validateAsPubKey` vs. `validatePubKey(key)`. 

Does this make sense? What do you think?

adityasaky: Yes, that makes sense. Since you'd mentioned looking into validating it as `key.validate()` and letting it take care of itself, I was thinking about how to do the check for it being a private key. However, I'll do something like this instead: https://github.com/adityasaky/in-toto-golang/blob/add-validate/in_toto/model.go#L36 which we can reuse if necessary. This would also mean that the `Key` struct has no validation methods itself (using a receiver).

adityasaky:
Coming to the validation of fields: I'm currently assuming we're working off a schema stored as a map which identifies the fields and whether they're required or optional. Now, typically you'd expect this validation to occur when you load the metadata itself from a file, and unmarshal it. However, the rest of the validations are currently operating on `Layout` or `Link`, and not `Metablock`. Given that, should the validation for this be different? If we bundle this in with the other validation, the system could be potentially loading incorrect metadata without complaining, which may have other repercussions I'm uncertain about, especially if validations aren't immediately called.

I suppose it's also possible to call `validate` from `Metablock` when loading, and continue allowing the schema validation to live with the other validation methods.

Alternatively, we can go the non receiver route wrt validation, something @lukpueh suggested I look into, where have `validateLayout(l *Layout)` and other similar methods for the other metadata. That'd change the basic validation setup itself, and we wouldn't quite have to generalize anything.

lukpueh:
> ... working off a schema stored as a map  ...

This seems redundant to your `validate*` functions. E.g. why check if the unmarshaled json file has some value on the required `_type` field, when you already have a `validateType` function that asserts the specific value for that field? I think the load function should do:
1. Unmarshal json into corresponding structs, which already entails a basic form of validation due to the data types defined on the structs.
  1.B Error/warn if the json file has any fields not defined by the destination structs
2. Call `validate` on the loaded struct, which traverses all fields and sub-fields, calling their corresponding `validate*` functions.

> ... we can go the non receiver route ...

I just wanted you to spend some thoughts on the pros and cons of passing the struct as receiver or regular argument. I think both approaches equally lend itself to generalization. Also, use generalization with moderation. We don't need  a general-purpose validation library. The goal is to have an easy to read and DRY set of validation functions that is structured hierarchically, so that I can validate any of our used top-level and lower-level metadata structures with a single function call.

lukpueh:
Some more thoughts...
IMO it is fine if the validation function name makes it clear, what you want to validate, e.g. `validateMetablock` (link or layout), `validateLink`, `validateLayout`, `validateKey` (private or public), `validePrivateKey`, `validatePublicKey`, etc ...

adityasaky:
> why check if the unmarshaled json file has some value on the required _type field, when you already have a validateType function that asserts the specific value for that field

Are you referring to https://github.com/adityasaky/in-toto-golang/blob/add-validate/in_toto/model.go#L348? What that's doing is checking if the loaded metadata is a Layout or Link and what struct to use. But I see your point, and we can do away with `validateType` unless we need it independently as a check at some other point, and not just when loading from file.

>  1.B Error/warn if the json file has any fields not defined by the destination structs

Right, for that though, we'd have to use [DisallowUnknownFields](https://golang.org/pkg/encoding/json/#Decoder.DisallowUnknownFields), moving away from `json.Unmarshal`, or use a slice/map of allowed fields and check manually.

> IMO it is fine if the validation function name makes it clear, what you want to validate, e.g. validateMetablock (link or layout), validateLink, validateLayout, validateKey (private or public), validePrivateKey, validatePublicKey, etc ...

I think this may be the best way forward if we don't want want multiple `validate` functions.

adityasaky:
I copied some of the individual processes into `validateLayout` etc. Does that seem better, @lukpueh?

in_toto/model.go
```diff
@@ -31,6 +33,24 @@ type Key struct {
 	Scheme              string   `json:"scheme"`
 }
 
+func validateKeyIdFormat(keyId string) error {
+	keyIdFormatCheck, err := regexp.MatchString("[a-fA-F0-9]+", keyId)
+	if err != nil {
+		return fmt.Errorf("unable to check if key ID has valid format")
+	}
```
lukpueh: You can omit that `err != nil` check. It [only returns an error if it can't compile the pattern](https://golang.org/src/regexp/regexp.go?s=15411:15479#L481), which you control.

in_toto/model.go
```diff
@@ -31,6 +33,24 @@ type Key struct {
 	Scheme              string   `json:"scheme"`
 }
 
+func validateKeyIdFormat(keyId string) error {
```
lukpueh: Minor nit: I'd omit the `Format`-suffix, because you don't have it anywhere else.

adityasaky:
https://golang.org/pkg/reflect/#Value.MapRange

I'm using this which was added in go 1.12. Travis CI and AppVeyor seem to be using an older version - Travis specifically is on 1.11. The tests are passing locally. Should I not use MapRange?

lukpueh:
Re go 1.12 on appveyor: https://github.com/appveyor/ci/issues/2875

in_toto/model.go
```diff
@@ -31,6 +35,24 @@ type Key struct {
...
 
+func validateKeyId(keyId string) error {
+	keyIdFormatCheck, _ := regexp.MatchString("^[a-fA-F0-9]+$", keyId)
+	if !keyIdFormatCheck {
+		return fmt.Errorf("'Key ID' has invalid format")
+	}
+	return nil
+}
+
+func validatePubKey(key Key) error {
```
lukpueh: Shouldn't `validatePubKey` call `validateKeyId` on its `KeyId`?

lukpueh: Please also note that key schemas in the reference implementation adds a bunch of more restrictions.

Currently we only support rsa keys, so it's probably worth having at least a `validatePublicRSAKey` (see `securesystemslib.formats.RSAKEY_SCHEMA`)

in_toto/model.go
```diff
@@ -31,6 +35,24 @@ type Key struct {
...
+	}
+	if key.KeyVal.Public == "" {
+		return fmt.Errorf("public key cannot be empty")
+	}
+	return nil
+}
+
 /*
 Signature represents a generic in-toto signature that contains the identifier
 of the Key, which was used to create the signature and the signature data.  The
```
lukpueh: What about `validateSignature`? See `securesystemslib.formats.SIGNATURE_SCHEMA`.

in_toto/model.go
```diff
@@ -57,6 +79,42 @@ type Link struct {
 	Environment map[string]interface{} `json:"environment"`
 }
 
+func validateMaterialOrProductHash(value string) error {
```
lukpueh: IMO this could be a generic `_validateHexString` function that may be used for keyids, signature value, artifact hashes, etc., whereas...

in_toto/model.go
```diff
@@ -57,6 +79,42 @@ type Link struct {
...
+	}
+	return nil
+}
+
+func validateLink(link Link) error {
+	if link.Type != "link" {
+		return fmt.Errorf("invalid type for link: should be 'link'")
+	}
+
+	for _, material := range link.Materials {
```
lukpueh: ... these two loops could be generalized into a single `validateArtifact` function. What do you think?

in_toto/model.go
```diff
@@ -112,6 +170,24 @@ type Step struct {
...
+		}
+	}
+	return nil
+}
+
+/*
+ISO8601DateSchema defines the format string of a timestamp following the
+ISO 8601 standard.
+*/
+const ISO8601DateSchema = "2006-01-02T15:04:05Z"
```
lukpueh: OT: Can you replace `RFC3339`  with `ISO8601DateSchema` in  `verifylib.VerifyLayoutExpiration`. I think it's okay to sneak this into this PR.

in_toto/model.go
```diff
@@ -150,6 +226,51 @@ func (l *Layout) InspectAsInterfaceSlice() []interface{} {
 	return inspectionsI
 }
 
+func validateLayout(layout Layout) error {
+	if layout.Type != "layout" {
+		return fmt.Errorf("invalid Type value for layout: should be 'layout'")
+	}
+
+	if _, err := time.Parse(ISO8601DateSchema, layout.Expires); err != nil {
```
lukpueh: Just to make sure,  will this check blow up if `layout.Expires` is a slightly different flavor of `ISO8601`, e.g. with milliseconds or utc offset?

in_toto/model.go
```diff
@@ -220,16 +341,58 @@ func (mb *Metablock) Load(path string) error {
...
+
+		for i := 0; i < attributeCount; i++ {
+			allFields = append(allFields, reflection.Field(i).Tag.Get("json"))
+		}
+
+		for _, field := range allFields {
+			if _, ok := signed[field]; !ok {
+				return fmt.Errorf("required field %s missing", field)
+			}
+		}
```
lukpueh: IMO It should be enough to just call `validateLink` after having deserialized the json file. It should blow up if the any of the validated link's fields are invalid, which includes required fields being  empty (`nil`?).

lukpueh: Same goes for `validateLayout` below.

in_toto/model.go
```diff
@@ -150,6 +226,51 @@ func (l *Layout) InspectAsInterfaceSlice() []interface{} {
...
+		} else {
+			namesSeen[inspection.Name] = true
+		}
+	}
+	return nil
+}
+
 /*
 Metablock is a generic container for signable in-toto objects such as Layout
 or Link.  It has two fields, one that contains the signable object and one that
```
lukpueh: What about `validateMetablock`?

adityasaky: Added. Will add tests shortly.

in_toto/model.go
adityasaky: It's now being validated as part of `validateMetablock`, should that be broken off into `validateSignature` and then called there? Are there other places where `validateSignature` can come in handy?

in_toto/model.go
adityasaky: I do still need to add GPG Signature check though. I'll do that too, probably as a separate helper.

in_toto/model.go
adityasaky: You're suggesting moving the required field check to `validateLink` and `validateLayout` respectively? I think in the meeting we had last month to discuss validation, we planned to check for the fields before unmarshalling into the structure? I may be confused. It should be an easy fix though.

in_toto/model.go
lukpueh: Yes, please make it a separate function, even if it's only called inside `validateMetablock` for now.

We still need to decide which parts of an in-toto implementation should be part of a public API (see https://github.com/in-toto/in-toto/issues/277), but IMO it's likely that there will be API functions that take `Signature` structs as argument, and we should always validate all arguments in public facing functions.

in_toto/model.go
lukpueh: We only need to validate before unmarshalling, to catch undefined fields, since those would be silently discarded during unmarshalling (you already do this with `DisallowUnknownFields`).

But I don't see the benefit of also checking required fields upfront, if they are validated afterwards anyway. Or am I missing something? 

Here's a simple example
```go
type Foo struct {
	Bar string `json:"bar"`
}
func validateFoo(foo Foo) error {
	if foo.Bar != "this string" {
		return fmt.Errorf("invalid foo: Bar must be 'this string', got: %s", foo.Bar)
	}
}
```

Now if you call `validateFoo` after unmashalling any json into `Foo`, it will blow up if the json was not `{"bar": "this string"}`, including the case where `"bar"` wasn't even in the json.

in_toto/model.go
adityasaky: This has been generalized to `validateHexSchema` + `validateRSAPubKey` now.

in_toto/model.go
lukpueh: @adityasaky, just informed me that this approach won't work, because the json unmarshaller initializes missing fields to the type's default value ("", 0, False ...), which might be a valid value for a required field.

https://github.com/golang/go/issues/17163 has some workarounds and what @adityasaky is doing is actually even better. So please disregard my comment from above.

However, it might be worth to create a function that does the json attribute iterating, to not repeat it for links and layouts...

in_toto/model.go
```diff
@@ -41,6 +81,25 @@ type Signature struct {
...
 
+func validateSignature(signature Signature) error {
+	if err := validateHexString(signature.KeyId); err != nil {
+		return fmt.Errorf("keyid: %s", err.Error())
+	}
+	if err := validateHexString(signature.Sig); err != nil {
+		return fmt.Errorf("signature: %s", err.Error())
+	}
+	return nil
+}
```
lukpueh: For `validateSignature` and also `validatePubKey` and `validateRSAPubKey` it might be worth to add the keyid to each error message, to make troubleshooting easier.

in_toto/model.go
```diff
@@ -171,6 +317,32 @@ type Metablock struct {
...
+	if signed["_type"] == "link" {
+		var link Link
+		reflection = reflect.TypeOf(link)
+	} else if signed["_type"] == "layout" {
+		var layout Layout
+		reflection = reflect.TypeOf(layout)
+	} else {
+		return fmt.Errorf("The '_type' field of the 'signed' part of in-toto" +
+			" metadata must be one of 'link' or 'layout'")
+	}
```
lukpueh: I suggest to call `reflect.TypeOf` in `Load` below, where you already have`Link`/`Layout` switch, and pass it as an additional argument to `checkRequiredFields` inside the switch. What do you think?

in_toto/model.go
```diff
@@ -112,6 +200,24 @@ type Step struct {
...
 
+func validateStep(step Step) error {
+	if step.Type != "step" {
+		return fmt.Errorf("invalid Type value for step: should be 'step'")
+	}
+	for _, keyId := range step.PubKeys {
+		if err := validateHexString(keyId); err != nil {
+			return fmt.Errorf("keyid: %s", err.Error())
+		}
+	}
```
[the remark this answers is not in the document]
lukpueh: Speaking of names should we add a "non-empty string" check for Name?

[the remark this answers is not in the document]
lukpueh: Btw. what about `validateInspection`? It could also use a `Name` and `Type` check.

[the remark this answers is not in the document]
lukpueh: And what about validation of artifact rules? You can use `UnpackRule`, which returns an error if the rule is not compliant.

You can also create a `validateSupplyChainItem` that is called by `validateStep` and `validateInspection` to summarize `Name`, `ExpectedMaterials` and `ExpectedProducts` checks.

in_toto/model.go
```diff
@@ -57,6 +116,35 @@ type Link struct {
 	Environment map[string]interface{} `json:"environment"`
 }
 
+func validateArtifacts(artifacts map[string]interface{}) error {
+	for _, artifact := range artifacts {
+		artifactValue := reflect.ValueOf(artifact).MapRange()
+		for artifactValue.Next() {
+			value := artifactValue.Value().Interface().(string)
+			if err := validateHexString(value); err != nil {
+				return fmt.Errorf("hash value: %s", err.Error())
```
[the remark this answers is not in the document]
adityasaky: The invalid key is part of `err.Error()`. That's what `validateHexString` returns. Added corresponding artifact name though.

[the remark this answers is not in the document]
lukpueh: Nope, I don't think so. By "hash key" I meant e.g. `"sha256"`. The error string returned by `validateHexString` only contains the "hash value", e.g. `"52947cb78b91ad55..."`.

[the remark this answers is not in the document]
adityasaky: Ah, understood. I'll add that.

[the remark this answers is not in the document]
lukpueh: I admit, currently we only have `"sha256"` but your `validateArtifacts` function is already written to support arbitrary artifact  maps, so we might as well let the user know which element in the map blew up. ;)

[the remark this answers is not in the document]
adityasaky: Done.

in_toto/model.go
adityasaky: The incorrect `keyid`? That's part of `err.Error()`.

```func validateHexString(str string) error {
	formatCheck, _ := regexp.MatchString("^[a-fA-F0-9]+$", str)
	if !formatCheck {
		return fmt.Errorf("'%s' is not a valid hex string", str)
	}
	return nil
}```

Comment (lukpueh) [reply in_toto/model.go]: Yes but only if the `KeyId` is the fail reason. What I'm suggesting is to add the `KeyId` to all error messages, where the KeyId helps to identify the piece of data, even if the `KeyId` is not the fail reason.

Comment (adityasaky) [reply in_toto/model.go]: Gotcha.

File: in_toto/model.go (hunk 18 above)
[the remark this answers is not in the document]
Comment (adityasaky) [reply in_toto/model.go]: `ExpectedMaterials` and `ExpectedProducts` only have type checks in the reference implementation, which can be dropped explicitly here.

File: in_toto/model.go (hunk 17 above)
Comment (adityasaky) [reply in_toto/model.go]: Unfortunately, I'm unable to pass a value of type `reflect.rtype` because it's taken to be private? Seems like it's the case because it's not capitalized in the `reflect` package.

**Comment by adityasaky** [comment]
Apart from passing `reflect` and validating `ExpectedProducts` and `ExpectedMaterials`, I think I've addressed everything else.

File: in_toto/model.go (hunk 17 above)
Comment (lukpueh) [reply in_toto/model.go]: Why not pass `reflect.Type`...?
```golang
func checkRequiredFields(signed map[string]interface{}, reflection reflect.Type) error {
  attributeCount := reflection.NumField()
  //...
```

And you call in `Load` below:
```golang
// ...
if signed["_type"] == "link" {
  var link Link
  if err := checkRequiredFields(signed, reflect.TypeOf(link)); err != nil {
    return err
  }
  // ...  repeat for `signed["_type"] == "layout" `
```

adityasaky: I didn't realize .Type was the same thing, apologies.

lukpueh: No worries. ;)

Issue #28, opened 2019-03-08
Support CREATE, DELETE, MODIFY


SantiagoTorres:
**Description of issue or feature request**:
The current set of rules supported by this implementation doesn't include `CREATE`, `DELETE`, and `MODIFY`. 

**Current behavior**: These rules aren't parsed accordingly (which leads to failure).

**Expected behavior**: These rules should be processed according to the spefcification.

PR #34, opened 2019-03-13
Add Set struct and set operations

lukpueh:
**Fixes issue #**:
pre-requisite for #28.

Golang does not support a set data structure (due to missing generics). This PR adds a simple set data structure for string elements and receivers for common set operations `has`, `add`, `remove`, `intersection` and `difference`, plus two handy receivers for pattern matching (filter) and casting a set to a string array (slice).

*Teaser*: 
The newly added `Filter` and `Intersection` functions will replace `verifylib`'s `FnFilter` and `Subtract` functions, when working on #28.

requirements**:

### DIFF
--- a/in_toto/util.go
+++ b/in_toto/util.go
@@ -1,0 +1,117 @@
+package in_toto
+
+import (
+	"fmt"
+	"path/filepath"
+)
+
+/*
+Set represents a data structure for set operations. See `NewSet` for how to
+create a Set, and available Set receivers for useful set operations.
+
+Under the hood Set aliases map[string]struct{}, where the map keys are the set
+elements and the map values are a memory-efficient way of storing the keys.
+*/
+type Set map[string]struct{}
+
+/*
+NewSet creates a new Set, assigns it the optionally passed variadic string
+elements, and returns it.
+*/
+func NewSet(elems ...string) Set {
+	var s Set
+	s = make(map[string]struct{})
+	for _, elem := range elems {
+		s.Add(elem)
+	}
+	return s
+}
+
+/*
+Has returns True if the passed string is member of the set on which it was
+called and False otherwise.
+*/
+func (s Set) Has(elem string) bool {
+	_, ok := s[elem]
+	return ok
+}
+
+/*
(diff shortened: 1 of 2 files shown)


in_toto/util.go
```diff
@@ -0,0 +1,117 @@
+package in_toto
+
+import (
+	"fmt"
+	"path/filepath"
+)
+
+/*
+Set represents a data structure for set operations. See `NewSet` for how to
+create a Set, and available Set receivers for useful set operations.
```
SantiagoTorres: My first dumb question is: hasn't anyone else already implemented such a fundamental data structure?

SantiagoTorres:
I'm a little confused by this PR. It seems to just be a facade over the existing set/map struct? What's the motivation for this? (sorry I may just be too much of a go illiterate)

in_toto/util.go
lukpueh: Yes and no, see https://stackoverflow.com/questions/34018908/golang-why-dont-we-have-a-set-datastructure. It didn't seem like there was "the one set implementation".

SantiagoTorres: From:  https://stackoverflow.com/questions/34018908/golang-why-dont-we-have-a-set-datastructure

> Partly, because Go doesn't have generics 

Okay, I laughed. Let me give this a go.

lukpueh:
I am adding this, because artifact rule processing relies heavily on set operations. So far we only needed *difference*, for which I implemented a quick and very dirty O(n**2) solution (see verifylib.Subtract). But with #28 we will also need *intersection*, so I thought why not add a mini set library.

lukpueh:
Here's a sneak preview of how the Set struct will be used in `VerifyArtifacts`:  https://github.com/lukpueh/in-toto-golang/commit/8719857b46ac4e572a948ac18a8694adedf40c4f 

I find it nicer to add something to the artifacts "queue" by writing  `artifacts_queue.Add(artifact)` instead of `artifacts_queue[artifact] = struct{}{}`.

Does this make sense? (I'll submit a follow-up PR tomorrow)

SantiagoTorres:
It does, thanks for the clarification!

PR #35, opened 2019-03-14
Support CREATE, DELETE, MODIFY rules

### LINKED ISSUE #28: Support CREATE, DELETE, MODIFY

lukpueh:
**Fixes issue #**:
Closes #28 
Includes changes PRed in #34 (either review and merge #34 first, or only review here)

Add support for missing rule types CREATE, DELETE and MODIFY. 
Rule verification is largely based on set operations using the mini set library added with #34.
See commits and commit message for details, most notably b9ea81c0d6bb130ef3d77ff16c7bf046ee763257.

requirements**:

### DIFF
--- a/in_toto/verifylib.go
+++ b/in_toto/verifylib.go
@@ -131,13 +131,14 @@ passed items (step or inspection) to enforce and authorize artifacts (materials
 or products) reported by the corresponding link and to guarantee that
 artifacts are linked together across links.  In the beginning all artifacts are
 placed in a queue according to their type.  If an artifact gets consumed by a
-rule it is removed from the queue.  An artifact can only be consumed once by
-one set of rules.
+rule it is removed from the queue.  An artifact can only be consumed once in
+the course of processing the set of rules in ExpectedMaterials or
+ExpectedProducts.
 
-Rules of type MATCH, ALLOW and DISALLOW are supported.
+Rules of type MATCH, ALLOW, CREATE, DELETE, MODIFY and DISALLOW are supported.
 
-MATCH and ALLOW remove artifacts from the corresponding queues on success, and
-leave the queue unchanged on failure.  Hence, it is left to a subsequent
+All rules except for DISALLOW consume queued artifacts on success, and
+leave the queue unchanged on failure.  Hence, it is left to a terminal
 DISALLOW rule to fail overall verification, if artifacts are left in the queue
 that should have been consumed by preceding rules.
 */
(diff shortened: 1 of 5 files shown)


SantiagoTorres:
for some reason I think there's a creep from #34 even after I merged it. Could you rebase please?

in_toto/verifylib.go
```diff
@@ -201,78 +163,131 @@ func VerifyArtifacts(items []interface{},
...
 
-		default: // Something's wrong
+		default: // Something wrong
 			return fmt.Errorf("VerifyArtifacts received an item of invalid type,"+
 				" elements of passed slice 'items' must be one of 'Step' or"+
 				" 'Inspection', got: '%s'", reflect.TypeOf(item))
 		}
+
+		// Use the item's name to extract the corresponding link
 		srcLinkMb, exists := itemsMetadata[itemName]
```
SantiagoTorres: Silly question, what does `Mb` stand for in this context? I keep on thinking of MetaBlock, but I don't think this is quite right?

in_toto/verifylib.go
```diff
@@ -201,78 +163,131 @@ func VerifyArtifacts(items []interface{},
...
+			// on artifacts in that queue.  If a rule consumes an artifact (i.e. can
+			// be applied successfully), the artifact is removed from the queue. By
 			// applying a DISALLOW rule eventually, verification may return an error,
 			// if the rule matches any artifacts in the queue that should have been
 			// consumed earlier.
-			var queue []string
-			for name, _ := range artifacts {
-				queue = append(queue, name)
-			}
+			queue := verificationData["artifactPaths"].(Set)
```
SantiagoTorres: Curious, why do you need the parentheses here?

in_toto/verifylib.go
lukpueh: Yep, Mb stands for metablock:
https://github.com/in-toto/in-toto-golang/blob/32abde0de57f3f002f55ac6319fe284cae384715/in_toto/verifylib.go#L64

in_toto/verifylib.go
SantiagoTorres: Oh wow this syntax is horrible. Ok, thanks!

lukpueh:
Thanks for the review, @SantiagoTorres!

Issue #32, opened 2019-03-08
Symbolic link walking


SantiagoTorres:
**Description of issue or feature request**: in-toto golang and the python/java implementations should have consistent behaviors when running into symbolic links

**Current behavior**: The current behavior when runlib runs into a symlink is currently not defined.

**Expected behavior**: [In-toto should dereference the symlinks and track them in the same way the python implementation does](https://github.com/in-toto/in-toto/blob/develop/in_toto/runlib.py#L162)

darshildpatel:
I am taking a look at the issue now.

darshildpatel:
While resolving this issue, I was able to resolve the symlinks that are present in the same directories. What this means is the code is able to resolve the symlinks which have absolute paths in there creation or have a relative path starting from the same directory. Currently am facing problems with resolving the symlinks which have relative paths in symlinks creation.  Please suggest as to what approach should be taken.

SantiagoTorres:
Hmm, care to share a MWE of the issue? I think it'd be easier to look at it.

Does this behavior happen with the python reference implementation too?

lukpueh:
I agree with @SantiagoTorres, we need more details about the problems you're facing in order to help you. Also, have you looked at [`fastwalk`](https://godoc.org/golang.org/x/tools/internal/fastwalk) from [Go tools](https://github.com/golang/tools)? It seems to support symlinks, with the following caveat: 
> It is the walkFn's responsibility to prevent fastWalk from going into symlink cycles.

darshildpatel:
@SantiagoTorres I have committed the clean recursive solution to my branch here https://github.com/darshildpatel/in-toto-golang/tree/add-symlinks-support/in_toto . Let me know if that looks fine. I would add more test cases and an extra variable for the infinite recursion problem. 

@lukpueh I am not able to use the package fastWalk from Go tools. It seems we are not able to use word "internal" in the import statement which hinders us from using that package directly in the code. Would you have some suggestions for this ?

lukpueh:
Thanks for looking into this, @darshildpatel! From what I read (e.g. https://github.com/golang/go/issues/4759) the go folks seem to not want to provide a public function, because it is difficult to get it right and prevent infinite loops.

This reluctancy among the experts makes me skeptical if it is a good idea to implement it by ourselves? @SantiagoTorres, what do you think?

[...]

PR #37, opened 2019-03-22
Add symlinks support

darshildpatel:

**Fixes issue #**: #32 

**Description of pull request**: Handles the directory symlinks. The support for recursive symlinks is WIP.

requirements**:


JustinCappos:
Here be dragons!  We need to really review this carefully with all of the race conditions related to symlinks / dirs in mind.

shibumi:
@lukpueh I think It makes sense to touch this PR here as well, when I have finished the exclude and error handling one. Following symlinks is very useful for the actual run implementation, that we are planning.

shibumi:
I had a closer look on this PR and I think we can use `filepath`'s `EvalSymlink()` method instead of checking for symlinks via the `os` module:

https://golang.org/pkg/path/filepath/#EvalSymlinks

This PR also diverges with the current master a lot, I think a complete rewrite would be necessary :S

shibumi:
@darshildpatel can you explain why you didn't use `filepath.EvalSymLink()` here?
I am not sure if you used os.ReadLink() on purpose for validating all "steps" in a symlink or if you just don't know about EvalSymLink().

shibumi:
@lukpueh I am not sure about the behavior of this PR. Would be nice if you could tell me what we actually want here...

Let's imagine we have the following symlink chain:

**/tmp/test3 -> /tmp/test2 -> /tmp/test1**

Then `filepath.EvalSymlinks("/tmp/test3")` will output: `/tmp/test1`
If we use `os.Readlink("/tmp/test3")` we will have output: `test2`.

Right now I think `filepath.EvalSymlinks` is the better solution here, because we wouldn't rely on a recursion for solving this issue + it's less error-prone. What do you think?

lukpueh:
> Right now I think filepath.EvalSymlinks is the better solution here, because we wouldn't rely on a recursion for solving this issue + it's less error-prone. What do you think?

Apologies for not replying earlier. I think either approach are fine as long as it is documented. It looks like the recursion argument no longer holds true, because you do use a recursive approach in #55 anyways, right?

Regardless, we still need to decide if link metadata stores the path of the source or the target of a symlink. Did you have a chance to look at the python reference implementation and see how it's done there?

shibumi:
I think we can close this PR. Thanks for your effort @darshildpatel, your PR was a key to our new symlink support :heart:

shibumi:
> Regardless, we still need to decide if link metadata stores the path of the source or the target of a symlink. Did you have a chance to look at the python reference implementation and see how it's done there?

uff.. just saw this now.. I will open a new issue for this, so that we keep track of it.

lukpueh:
> uff.. just saw this now.. I will open a new issue for this, so that we keep track of it.

Thanks! I forgot to cross-check too. :)

lukpueh:
Thanks for the solid spadework, @darshildpatel!

PR #38, opened 2019-04-02
Add parameter substitution

### LINKED ISSUE #29: Support parameter substitution
Parameter substitution is not supported by this implementation.

**Current behavior**: Parameter substitution is not performed

**Expected behavior**: This implementation should provide a similar interface for integrators to perform parameter substitution.

adityasaky:

Adds parameter substitution using a map of parameters and their corresponding values. Fixes #29.

requirements**:

### DIFF
--- a/in_toto/verifylib.go
+++ b/in_toto/verifylib.go
@@ -654,6 +654,16 @@ func substituteParametersInSliceOfSlices(replacer *strings.Replacer,
 	return newSlice
 }
 
+/*
+SubstituteParameters performs parameter substitution in steps and inspections
+in the following fields:
+- Expected Materials and Expected Products of both
+- Run of inspections
+- Expected Command of steps
+The substitution marker is '{}' and the keyword within the braces are replaced
+by a value found in the substitution map passed, parameterDictionary. The
+layout with parameters substituted is returned to the calling function.
+*/
 func SubstituteParameters(layout Layout,
 	parameterDictionary map[string]string) (Layout, error) {
 
--- a/in_toto/verifylib.go
+++ b/in_toto/verifylib.go
@@ -684,10 +684,10 @@ func SubstituteParameters(layout Layout,
 	for _, inspection := range layout.Inspect {
 		inspection.ExpectedMaterials =
 			substituteParametersInSliceOfSlices(replacer,
-			inspection.ExpectedMaterials)
+				inspection.ExpectedMaterials)
 		inspection.ExpectedProducts =
 			substituteParametersInSliceOfSlices(replacer,
-			inspection.ExpectedProducts)
+				inspection.ExpectedProducts)
 		inspection.Run = substituteParamatersInSlice(replacer, inspection.Run)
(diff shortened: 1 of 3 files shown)


adityasaky:
For reference, this needs:

- addition of tests
- call in `InTotoVerify` (also has to include a flag as parameter substitution is optional)
- Improved error handling

adityasaky:
I initially did have the replacer in a helper function but then moved it into each instance as we do have a non list of list also. I also used the reference implementation for help there.

lukpueh:
What's the "non list of list", @adityasaky? AFAICS, you run you replacer on  lists of strings https://github.com/in-toto/in-toto-golang/blob/be4946db3e01e2999431c9568c8034f0bb2a08a2/in_toto/model.go#L110
https://github.com/in-toto/in-toto-golang/blob/be4946db3e01e2999431c9568c8034f0bb2a08a2/in_toto/model.go#L95)
and lists of lists of strings
https://github.com/in-toto/in-toto-golang/blob/be4946db3e01e2999431c9568c8034f0bb2a08a2/in_toto/model.go#L82-L83

I think a helper would be beneficial, because the reduced clutter makes it easier to see which layout fields are subjected to parameter substitution. It's not a blocker though. ;)

in_toto/verifylib.go
```diff
@@ -679,7 +752,12 @@ func InTotoVerify(layoutMb Metablock, layoutKeys map[string]Key,
 		return summaryLink, err
 	}
 
-	// TODO: Substitute parameters
+	if len(parameterDictionary) > 0 {
```
lukpueh: Minor nit: I think this check could be performed inside `SubstituteParameters`. It would make `InTotoVerify` slimmer. :)

in_toto/verifylib.go
```diff
@@ -661,7 +733,8 @@ NOTE: Parameter substitution, artifact rules of type "create", "modify"
 and "delete" are currently not supported.
```
lukpueh: Please also update `InTotoVerify`'s docstring (the verification routine part).

in_toto/verifylib.go
```diff
@@ -633,6 +634,77 @@ func VerifySublayouts(layout Layout,
...
+func SubstituteParameters(layout Layout,
+	parameterDictionary map[string]string) (Layout, error) {
+
+	parameters := make([]string, 0)
+
+	for parameter, value := range parameterDictionary {
+		parameterFormatCheck, _ := regexp.MatchString("^[a-zA-Z0-9_-]+$",
+			parameter)
+		if !parameterFormatCheck {
+			return layout, fmt.Errorf("invalid format for parameter")
```
lukpueh: Can you add a test for this?

lukpueh:
Thanks for addressing my comments, @adityasaky. +1 for the acronym ITV ;) 

And what a lovely coincidence that AppVeyor has added Go 1.12 support just a couple of hours ago. Merging...

Issue #33, opened 2019-03-08
Exclude patterns


SantiagoTorres:
**Description of issue or feature request**: The python implementation has a nice frontend feature to ignore artifacts a-la `.gitignore` we should port that.

**Current behavior**: This feature is not implemented, so runlib captures all the materials in the directory.

**Expected behavior**: It should be possible to specify patterns to ignore when collecting file integrity.

ayush159:
I have started working on this issue. I had a question regarding this:

In the python implementation, pathspec package is used. I have found a similar implementation of gitignore for GO here -> https://github.com/xanzy/go-pathspec . Can we use this package? 

I have pushed a partially working fix of this issue on my branch (https://github.com/ayush159/in-toto-golang/tree/master/in_toto) using this package.

SantiagoTorres:
Hi @ayush159 !

I think it's fine if you want to include pathspec into the implementation, although I may want to take a look at this pkg more closely to make sure the license and everything doesn't clash with what we're doing. What do you think @lukpueh?

I'm looking at your fix, and it looks like it's going in the right direction, but you may want to simplify the code a little bit (those big if/else blocks could probably be replaced with guard statements and whatnot).

lukpueh:
License compatibility seems fine. What worries me more is the low repo activity (most recent commit in 2014). Not sure if that's a problem though. Have you looked for alternatives? What about the 
[`gitignore`](https://godoc.org/gopkg.in/src-d/go-git.v4/plumbing/format/gitignore) subpackage from [`go-git`](https://github.com/src-d/go-git)?

lukpueh:
I think this feature was fully implemented with #51. But before we close here, let's also remove the corresponding [todo item from the repo README](https://github.com/in-toto/in-toto-golang/blame/f3f615b900aa4798e9f0558f2f90c37884d0345f/README.md#L31)

PR #41, opened 2019-04-12
Exclude_pattern fix (Issue #33)

ayush159:

**Fixes issue #33 **:

Used go-git to implement gitignore style pattern parsing. Changed the signature of RecordArtifacts function in runlib.go . Added exclude pattern in method signature. If exclude patterns not available, pass nil.

requirements**:


SantiagoTorres:
@lukpueh, I suspect this is failing on appveyor due to go being 1.11. Should we bump it to 1.12 in all the platforms?

adityasaky:
If just bumping the version isn't enough: https://github.com/in-toto/in-toto-golang/pull/25#issuecomment-480264294

When I faced this issue though, both Travis CI and AppVeyor was failing.

lukpueh:
Thanks for cross-referencing, @adityasaky! The corresponding AppVeyor GitHub issues also provides a temporary workaround. Any volunteers? :)

ayush159:
I was following the referenced AppVeyor issue, and have found a possible fix here
https://github.com/go-ggz/ggz/issues/80#issuecomment-482928898

If this fix looks okay, I can try this with my pull request and we can see if the build works? Though I am not sure if it's the best solution because this will reinstall Go on every build.

lukpueh:
@ayush159, please go ahead. Would mind submitting a separate PR for the fix? @adityasaky's #25 would also benefit from it.

lukpueh:
@ayush159, FYI AppVeyor has just deployed go 1.12. So, a workaround is no longer required.

in_toto/runlib.go
```diff
@@ -76,12 +91,25 @@ func RecordArtifacts(paths []string) (map[string]interface{}, error) {
 				if info.IsDir() {
 					return nil
 				}
-				artifact, err := RecordArtifact(path)
-				// Abort if artifact can't be recorded, e.g. due to file permissions
-				if err != nil {
-					return err
+				if exclude_patterns != nil {
```
SantiagoTorres: Could you please encapsulate this inside a function to improve grokability? I'm also wondering if we can do something to avoid this arrowhead pattern. Probably having the function behave like a filter. I also suspect you can restructure this to remove some of the repeated code.

ayush159: Ok, I'll work on these changes. Also, it seems that updating the Go version on AppVeyor wasn't enough to resolve the build fail issue. It's still failing with the same reason on AppVeyor. Though, the build is passing on Travis with Go 1.12.

shibumi:
what's the status on this PR here? Shall I adopt it? Do you want to continue working on it @ayush159 ?

lukpueh:
Looks like this has been stale for a year. I think it's fine for you to give it a shot, @shibumi. 💯for being pro-active! :)

lukpueh:
FYI we recently discovered a bug in the exclude pattern feature of the reference implementation. When working on this, I suggest adding a test case that would catch a similar bug here.

shibumi:
@lukpueh sorry, just saw your comment 4 days later. Of course I can have a look on this one, too :)

lukpueh:
Superseded by #53. Many thanks for the spadework, @ayush159; and for taking over, @shibumi! Closing here...

Issue #4, opened 2018-11-12
Update user/developer feedback


lukpueh:
- Review go logging libraries and find a good way to handle error, warning, info, debug messages
- Add/enhance log messages for different message types/levels, see @aaaaalbert's https://github.com/in-toto/in-toto-golang/pull/1#discussion_r229420766
- Use appropriate standard streams, see @SantiagoTorres's https://github.com/in-toto/in-toto-golang/pull/1#discussion_r232060444

alanssitis:
Is it worth revisiting this now that `slog` has finally reached a maturity I am comfortable with?

adityasaky:
I'd recommend working on any new in-toto-golang issues in conjunction with Go consolidation, to ensure we still want to add the feature in question.

PR #43, opened 2019-05-31
Refine validation

lukpueh:
**Fixes issue #**:
Post-merge refinement of @adityasaky's #25.

Simplifies json required field validation, adds artifact rule validation function and some more tests (with TDT setup).

requirements**:

### DIFF
--- a/in_toto/model.go
+++ b/in_toto/model.go
@@ -203,6 +203,17 @@ type SupplyChainItem struct {
 }
 
 /*
+validateArtifactRule calls UnpackRule to validate that the passed rule conforms
+with any of the available rule formats.
+*/
+func validateArtifactRule(rule []string) error {
+	if _, err := UnpackRule(rule); err != nil {
+		return err
+	}
+	return nil
+}
+
+/*
 validateSupplyChainItem is used to validate the common elements found in both
 steps and inspections. Here, the function primarily ensures that the name of
 a supply chain item isn't empty.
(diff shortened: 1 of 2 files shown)


lukpueh:
IMO the validation error messages could be refined some more, to always show what object is invalid, which part of it, and what's the correct value. We can revisit with  #4.

in_toto/model.go
```diff
@@ -211,6 +222,16 @@ func validateSupplyChainItem(supplyChainItem SupplyChainItem) error {
 	if supplyChainItem.Name == "" {
 		return fmt.Errorf("name cannot be empty")
 	}
+	for _, rule := range supplyChainItem.ExpectedMaterials {
```
adityasaky: Previously, we discussed having a helper for iterating and verifying signatures. We also do have something similar in paramsub, though that was messier.

https://github.com/in-toto/in-toto-golang/blob/0e8eb5b8845baffc91cd4b494772b2b77c4ddf0f/in_toto/model.go#L98-L120

https://github.com/in-toto/in-toto-golang/blob/7e0a1a42ec0c1d5d894bd576fbd8fed1cb01da12/in_toto/verifylib.go#L637-L655

Think that's relevant here?

in_toto/model.go
```diff
@@ -202,6 +202,17 @@ type SupplyChainItem struct {
...
+validateArtifactRule calls UnpackRule to validate that the passed rule conforms
+with any of the available rule formats.
+*/
+func validateArtifactRule(rule []string) error {
+	if _, err := UnpackRule(rule); err != nil {
+		return err
+	}
+	return nil
+}
+
```
adityasaky: Should there be an explicit `TestValidateArtifactRule` for this?

adityasaky: Should there be an explicit `TestValidateArtifactRule` for this?

in_toto/model.go
lukpueh: Not sure how it is relevant. Can you elaborate?

in_toto/model.go
lukpueh: IMO we're good here. It's only a wrapper for `UnpackRule`, which is extensively tested in `rulelib_test.go`. And it's covered by higher-level validate functions, such as `TestValidateSupplyChainItem`.

in_toto/model.go
adityasaky: Basically, I’m asking if there should be a verifyArtifactRules that in turn iterates and calls verifyArtifactRule, similar to how it’s setup for signatures etc elsewhere.

lukpueh: Done in 0766aa7

lukpueh:
Thanks for your review, @adityasaky. Let me know if it's good to merge. :)

adityasaky:
Looks good to me.

PR #45, opened 2019-08-15
Add test command to Travis configuration

radu-matei:
The current Travis config does not actually test the package.
This fix adds the `go test ./...` command to the Travis configuration.

Signed-off-by: Radu M <root@radu.sh>

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,5 +7,6 @@ before_install:
   - go get github.com/mattn/goveralls
 
 script:
+  - go test ./...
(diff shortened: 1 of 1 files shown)


lukpueh:
Thanks for the patch, @radu-matei!

We are using `goveralls` to run our test suite and publish the coverage to `goveralls.io`, thus there *["is no need to call  `go test` separately"](https://github.com/mattn/goveralls#continuous-integration)*.

PR #48, opened 2019-10-22
in_toto/{keylib,model}: add ed25519 signing

SantiagoTorres:

**Fixes issue #**:

requirements**:

### DIFF
--- a/in_toto/keylib.go
+++ b/in_toto/keylib.go
@@ -2,7 +2,6 @@ package in_toto
 
 import (
 	"crypto"
-	"crypto/ed25519"
 	"crypto/rsa"
 	"crypto/sha256"
 	"crypto/x509"
@@ -10,6 +9,7 @@ import (
 	"encoding/json"
 	"encoding/pem"
 	"fmt"
+	"golang.org/x/crypto/ed25519"
 	"io/ioutil"
 	"os"
 	"reflect"
--- a/in_toto/keylib.go
+++ b/in_toto/keylib.go
@@ -156,8 +156,8 @@ func VerifySignature(key Key, sig Signature, data []byte) error {
 }
 
 /*
-ParseEd25519FromPrivateJSON parses and ed25519 private key from the json
-string. These ed25519 keys have the format as generated using in-toto-keygen:
+ParseEd25519FromPrivateJSON parses an ed25519 private key from the json string.
+These ed25519 keys have the format as generated using in-toto-keygen:
 {
   "keytype: "ed25519",
   "scheme": "ed25519",
@@ -194,11 +194,11 @@ func ParseEd25519FromPrivateJSON(JSONString string) (Key, error) {
 }
 
 /*
-generateEd25519Signature creates an ed25519 signature using the key and the
+GenerateEd25519Signature creates an ed25519 signature using the key and the
 signable buffer provided. It returns an error if the underlying signing library
 fails.
 */
-func generateEd25519Signature(signable []byte, key Key) (Signature, error) {
+func GenerateEd25519Signature(signable []byte, key Key) (Signature, error) {
 
 	var signature Signature
 
--- a/in_toto/keylib.go
+++ b/in_toto/keylib.go
@@ -155,21 +155,21 @@ func VerifySignature(key Key, sig Signature, data []byte) error {
 	return nil
 }
 
-/* Parse ed25519 from private json uses the passed json string to populate and
-* ed25519 key. These ed25519 keys have the format as generated using
-* in-toto-keygen:
-*  {
-    "keytype: "ed25519",
-    "scheme": "ed25519",
-    "keyid": ...
-    "keyid_hash_algorithms": [...]
-    "keyval": {
-        "public": "..." # 32 bytes
-        "private": "..." # 32 bytes
-    }
-   }
+/*
+ParseEd25519FromPrivateJSON parses and ed25519 private key from the json
+string. These ed25519 keys have the format as generated using in-toto-keygen:
+{
+  "keytype: "ed25519",
+  "scheme": "ed25519",
+  "keyid": ...
+  "keyid_hash_algorithms": [...]
+  "keyval": {
+    "public": "..." # 32 bytes
+    "private": "..." # 32 bytes
+  }
+}
 */
-func Parseed25519FromPrivateJSON(JSONString string) (Key, error) {
+func ParseEd25519FromPrivateJSON(JSONString string) (Key, error) {
 
 	var keyObj Key
 	err := json.Unmarshal([]uint8(JSONString), &keyObj)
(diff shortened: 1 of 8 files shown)


in_toto/model.go
```diff
@@ -300,3 +300,34 @@ func (mb *Metablock) VerifySignature(key Key) error {
...
+	// FIXME: we could be fancier about signature-generation using a dispatch
+	// table or something but for now let's just be explicit
+	// (also, lolnogenerics)
+	if key.KeyType == "ed25519" && key.Scheme == "ed25519" {
+		newSignature, err = generateEd25519Signature(dataCanonical, key)
+		if err != nil {
+			return err
+		}
+	} else {
+		return fmt.Errorf("This key type or signature scheme is not supported yet!")
```
lukpueh: Adding `key.KeyType` and `key.Scheme` to the error message would be quite fancy. Other than that I think this form of dispatching is fancy enough.

in_toto/keylib.go
```diff
@@ -152,3 +154,63 @@ func VerifySignature(key Key, sig Signature, data []byte) error {
...
+    "scheme": "ed25519",
+    "keyid": ...
+    "keyid_hash_algorithms": [...]
+    "keyval": {
+        "public": "..." # 32 bytes
+        "private": "..." # 32 bytes
+    }
+   }
+*/
+func Parseed25519FromPrivateJSON(JSONString string) (Key, error) {
```
lukpueh: It took me a few seconds to parse `Parseed...` (also because we deal with a `seed` below). Can we change this to `ParseEd...`?

in_toto/keylib.go
```diff
@@ -152,3 +154,63 @@ func VerifySignature(key Key, sig Signature, data []byte) error {
 
 	return nil
 }
+
+/* Parse ed25519 from private json uses the passed json string to populate and
```
lukpueh:
```suggestion
/* Parse ed25519 from private json uses the passed json string to populate an
```

lukpueh: Also, I think per convention go function docstrings start with the function name.

SantiagoTorres:
thanks, @lukpueh ! This should address most/all of your comments hopefully

SantiagoTorres:
Huh!? for some reason this stack doesn't have `crypto/ed25519` this builds locally with:
```
go version go1.13.3 linux/amd64
```

lukpueh:
Seems like the `Parseed` --> `ParseEd` refactor was not complete.

lukpueh:
Also, the go version in `go.mod` should probably be bumped.

lukpueh:
Seem like [appveyor/windows only has go 1.10](https://www.appveyor.com/docs/lang/go/#windows) :(

SantiagoTorres:
:[ that's EOL

in_toto/keylib.go
```diff
@@ -152,3 +154,63 @@ func VerifySignature(key Key, sig Signature, data []byte) error {
...
+
+	return keyObj, nil
+}
+
+/*
+generateEd25519Signature creates an ed25519 signature using the key and the
+signable buffer provided. It returns an error if the underlying signing library
+fails.
+*/
+func generateEd25519Signature(signable []byte, key Key) (Signature, error) {
```
lukpueh: We should probably also make this public, right? I.e.  `s/generateEd25519Signature/GenerateEd25519Signature/`.

in_toto/keylib_test.go
```diff
@@ -141,3 +141,68 @@ k7Gtvz/iYzaLrZv33cFWWTsEOqK1gKqigSqgW9T26wO9AgMBAAE=
 		}
 	}
 }
+
+func testParseEd25519FromPrivateJSON(t *testing.T) {
```
lukpueh: Judging from the coverage report, these tests didn't run. 
`s/test/Test/` seems to do the trick. :)

in_toto/keylib_test.go
```diff
@@ -141,3 +141,68 @@ k7Gtvz/iYzaLrZv33cFWWTsEOqK1gKqigSqgW9T26wO9AgMBAAE=
...
+	validKey := `{"keytype": "ed25519", "scheme": "ed25519", "keyid": "308e3f53523b632983a988b72a2e39c85fe8fc967116043ce51fa8d92a6aef64", "keyid_hash_algorithms": ["sha256", "sha512"], "keyval": {"public": "8f93f549eb4cca8dc2142fb655ba2d0955d1824f79474f354e38d6a359e9d440", "private": "861fd1b466cfc6f73f8ed630f99d8eda250421f0e3a6123fd5c311cc001bda49"}}`
+	_, err := ParseEd25519FromPrivateJSON(validKey)
+	if err != nil {
+		t.Errorf("ParseEd25519FromPrivateJSON returned (%s), expected no error",
+			err)
+	}
+
+}
+
+func testgenerateEd25519Signature(t *testing.T) {
```
lukpueh: `s/test/Test/`

in_toto/model_test.go
```diff
@@ -1128,3 +1128,54 @@ func TestValidateSupplyChainItem(t *testing.T) {
...
+	var mb Metablock
+	mb.Load("demo.layout.template")
+
+	key.LoadPublicKey("alice.pub")
+	err := mb.Sign(key)
+	if err == nil || !strings.Contains(err.Error(), "supported yet") {
+		t.Errorf("Metablock.Sign returned (%s), expected it to claim this "+
+			"key type/scheme is unsupported", err)
+
+	}
```
lukpueh: This fails because its type is non-ed25519 not because it is non-private, right?  Could you also add a test case for the latter, in order to return an error [here](https://coveralls.io/jobs/54864621/source_files/769976287#L642)?

in_toto/model_test.go
```diff
@@ -1128,3 +1128,54 @@ func TestValidateSupplyChainItem(t *testing.T) {
...
+	err = mb.Sign(key)
+	if err != nil {
+		t.Errorf("Metablock.Sign returned (%s), expected no error", err)
+	}
+
+	// note, there's a 2 because this template is already signed hehe
+	if len(mb.Signatures) != 2 {
+		t.Errorf("Expected a new signature to be appended, but got (%+v) (%d)",
+			mb.Signatures, len(mb.Signatures))
+	}
```
lukpueh: Have you tried verifying the signature? Or do you want to do this in a separate PR?

SantiagoTorres: I want to abstract verification functions a little bit in a separate PR, if that's ok with you :)

in_toto/model_test.go
SantiagoTorres: I did add a test [there](https://github.com/in-toto/in-toto-golang/pull/48/commits/cbe0e6106bd76fdd58d0309e28aed1029c64c8fc#diff-5b7f068c4ad310ffd7ed40b65016260eR1149), but it's unfortunately not triggered due to the order of the assertions. I may be missing something and you were thinking of a different way to trigger that line?

lukpueh:
@SantiagoTorres, I added the test I had in mind in f003324. I also played around a little bit with AppVeyor. I was wrong about the supported go version. It does [go up to 1.12.3](https://www.appveyor.com/docs/windows-images-software/#golang), but I didn't figure out why it can't find `golang.org/x/crypto/ed25519`.

Issue #49, opened 2019-10-24
canonical_json encoder coerces floats to ints


SantiagoTorres:
**Description of issue or feature request**:
It appears that our implementation of canonical_json is failing the spec, as it coerces float types to ints. The error was nailed down to [this line](https://github.com/in-toto/in-toto-golang/blob/master/in_toto/canonicaljson.go#L53-L59) by @radu-matei 

**Current behavior**:
The input object

```json
{
    "anumber": 1.2,
    "znumber": 2.3
}
```

outputs
`{"anumber":1,"znumber":2}`

**Expected behavior**:
The routine should fail, as do the rust and python implementations. Do, see this [snippet](https://bpaste.net/show/PXJJM)

lukpueh:
Thanks, @radu-matei and @SantiagoTorres for pitching this issue. #50 should fix it.

shibumi:
@SantiagoTorres I think we can close this one.

lukpueh:
Fixed in #50

PR #50, opened 2019-10-24
Fix encode canonical

lukpueh:
**Fixes issue #**:
#49 

Prior to this change any numeric values were converted to integers during canonicalization. This does not conform with the OLPC canonical JSON specification, which disallows floating point numbers.

This PR changes `EncodeCanonical` to decode any numeric values as `json.Number` (instead of the default `float64`), and `_encodeCanonical` to fail if any `json.Number` is not a whole number.

Note that the json decoder in `Metablock.load` in `models` is not changed, and will decode any numeric values from a loaded json file as `float64`. However, these numbers may still be successfully encoded with `EncodeCanonical` as long as they hold whole numbers (see [tests](https://github.com/lukpueh/in-toto-golang/blob/814100d385dabf507399ff183eba16e3071db142/in_toto/canonicaljson_test.go#L27)).

This PR also fixes an error handling bug in `_encodeCanonical`.

requirements**:

### DIFF
--- a/in_toto/canonicaljson.go
+++ b/in_toto/canonicaljson.go
@@ -8,7 +8,6 @@ import (
 	"reflect"
 	"regexp"
 	"sort"
-	"strconv"
 )
 
 /*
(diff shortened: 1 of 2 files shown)


SantiagoTorres:
Thanks, @lukpueh !

PR #53, opened 2020-06-01
shibumi/exclude pattern fix 41

shibumi:
Note: This PR is based on PR #41, I couldn't work on the branch directly, because I am not a project maintainer, so I checkout out PR #41 and opened a new PR instead. With this approach we can still list @ayush159 as author for the code he wrote.

**EDIT: I rewrote everything from scratch**

I am right now working on:

* adding documentation (The Apply... func is missing documentation. The func is being exported, so we should definitely add some docs for it)
* Removing unnecessary things or improving error messages. For example: 

https://github.com/in-toto/in-toto-golang/pull/41/files#diff-73e399e3c84826b077d43cc4cbc98e7eR103

Original PR description is below:

**Fixes issue #33**:

**Description of pull request:**
Used go-git to implement gitignore style pattern parsing. Changed the signature of RecordArtifacts function in runlib.go . Added exclude pattern in method signature. If exclude patterns not available, pass nil.

Please verify and check that the pull request fulfills the following
requirements:

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -2,4 +2,7 @@ module github.com/in-toto/in-toto-golang
 
 go 1.14
 
-require golang.org/x/crypto v0.0.0-20191029031824-8986dd9e96cf
+require (
+	github.com/shibumi/go-pathspec v1.1.0
+	golang.org/x/crypto v0.0.0-20191029031824-8986dd9e96cf
+)
--- a/go.mod
+++ b/go.mod
@@ -1,8 +1,8 @@
-module github.com/in-toto/in-toto-golang
+module github.com/shibumi/in-toto-golang
 
-go 1.14
+go 1.15
 
 require (
-	github.com/shibumi/go-pathspec v1.1.0
-	golang.org/x/crypto v0.0.0-20191029031824-8986dd9e96cf
+	github.com/shibumi/go-pathspec v1.2.0
+	golang.org/x/crypto v0.0.0-20201112155050-0c6587e931a9
 )
--- a/in_toto/runlib.go
+++ b/in_toto/runlib.go
@@ -93,23 +93,6 @@ func RecordArtifacts(paths []string, hashAlgorithms []string, gitignorePatterns 
 }
 
 /*
-applyGitIgnorePatterns includes or excludes file via gitignore-style pathspec pattern matching.
-*/
-func applyGitIgnorePatterns(paths []string, excludePatterns []string) ([]string, error) {
-	var newPaths []string
-	for _, path := range paths {
-		ignore, err := pathspec.GitIgnore(excludePatterns, path)
-		if err != nil {
-			return []string{}, err
-		}
-		if !ignore {
-			newPaths = append(newPaths, path)
-		}
-	}
-	return newPaths, nil
-}
-
-/*
 recordArtifacts walks through the passed slice of paths, traversing
 subdirectories, and calls RecordArtifact for each file. It returns a map in
 the following format:
@@ -129,12 +112,6 @@ return value is the error.
 */
 func recordArtifacts(paths []string, hashAlgorithms []string, gitignorePatterns []string) (map[string]interface{}, error) {
 	artifacts := make(map[string]interface{})
-	// This should match paths evaluated via symlinks too, because recordArtifacts works
-	// recursively.
-	paths, err := applyGitIgnorePatterns(paths, gitignorePatterns)
-	if err != nil {
-		return artifacts, err
-	}
 	for _, path := range paths {
 		err := filepath.Walk(path,
 			func(path string, info os.FileInfo, err error) error {
@@ -142,6 +119,16 @@ func recordArtifacts(paths []string, hashAlgorithms []string, gitignorePatterns 
 				// e.g. path does not exist
 				if err != nil {
 					return err
+				}
+				// We need to call pathspec.GitIgnore inside of our filepath.Walk, because otherwise
+				// we will not catch all paths. Just imagine a path like "." and a pattern like "*.pub".
+				// If we would call pathspec outside of the filepath.Walk this would not match.
+				ignore, err := pathspec.GitIgnore(gitignorePatterns, path)
+				if err != nil {
+					return err
+				}
+				if ignore {
+					return nil
 				}
 				// Don't hash directories
 				if info.IsDir() {
--- a/in_toto/runlib.go
+++ b/in_toto/runlib.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"errors"
 	"fmt"
+	"github.com/shibumi/go-pathspec"
 	"io/ioutil"
 	"os"
 	"os/exec"
@@ -83,12 +84,29 @@ the following format:
 If recording an artifact fails the first return value is nil and the second
 return value is the error.
 */
-func RecordArtifacts(paths []string, hashAlgorithms []string) (evalArtifacts map[string]interface{}, err error) {
+func RecordArtifacts(paths []string, hashAlgorithms []string, gitignorePatterns []string) (evalArtifacts map[string]interface{}, err error) {
 	// Make sure to initialize a fresh hashset for every RecordArtifacts call
 	visitedSymlinks = NewSet()
-	evalArtifacts, err = recordArtifacts(paths, hashAlgorithms)
+	evalArtifacts, err = recordArtifacts(paths, hashAlgorithms, gitignorePatterns)
 	// pass result and error through
 	return evalArtifacts, err
+}
+
+/*
+applyGitIgnorePatterns includes or excludes file via gitignore-style pathspec pattern matching.
+*/
+func applyGitIgnorePatterns(paths []string, excludePatterns []string) ([]string, error) {
+	var newPaths []string
+	for _, path := range paths {
+		ignore, err := pathspec.GitIgnore(excludePatterns, path)
+		if err != nil {
+			return []string{}, err
+		}
+		if !ignore {
+			newPaths = append(newPaths, path)
+		}
+	}
+	return newPaths, nil
 }
 
 /*
@@ -109,9 +127,14 @@ the following format:
 If recording an artifact fails the first return value is nil and the second
 return value is the error.
 */
-func recordArtifacts(paths []string, hashAlgorithms []string) (map[string]interface{}, error) {
+func recordArtifacts(paths []string, hashAlgorithms []string, gitignorePatterns []string) (map[string]interface{}, error) {
 	artifacts := make(map[string]interface{})
-	// NOTE: Walk cannot follow symlinks
+	// This should match paths evaluated via symlinks too, because recordArtifacts works
+	// recursively.
+	paths, err := applyGitIgnorePatterns(paths, gitignorePatterns)
+	if err != nil {
+		return artifacts, err
+	}
 	for _, path := range paths {
 		err := filepath.Walk(path,
 			func(path string, info os.FileInfo, err error) error {
@@ -148,7 +171,7 @@ func recordArtifacts(paths []string, hashAlgorithms []string) (map[string]interf
 					visitedSymlinks.Add(path)
 					// We recursively call RecordArtifacts() to follow
 					// the new path.
-					evalArtifacts, evalErr := recordArtifacts([]string{evalSym}, hashAlgorithms)
+					evalArtifacts, evalErr := recordArtifacts([]string{evalSym}, hashAlgorithms, gitignorePatterns)
 					if evalErr != nil {
 						return evalErr
 					}
@@ -256,9 +279,9 @@ Metablock object.  If command execution or artifact recording fails the first
 return value is an empty Metablock and the second return value is the error.
 */
 func InTotoRun(name string, materialPaths []string, productPaths []string,
-	cmdArgs []string, key Key, hashAlgorithms []string) (Metablock, error) {
+	cmdArgs []string, key Key, hashAlgorithms []string, gitignorePatterns []string) (Metablock, error) {
 	var linkMb Metablock
-	materials, err := RecordArtifacts(materialPaths, hashAlgorithms)
+	materials, err := RecordArtifacts(materialPaths, hashAlgorithms, gitignorePatterns)
 	if err != nil {
 		return linkMb, err
 	}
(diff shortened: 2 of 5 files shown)


shibumi:
@lukpueh If you see anything else I can fix in this code, let me know. We should also have a look on the AppVeyor pipeline.. it seems we are running into problems like in the original PR.

shibumi:
@lukpueh oh and I am confused about the `exclude_patterns` variable in the metablock. Is this `exclude_patterns` in the official in-toto specification? If not, we should add it. If yes, then we should probably rename it, because I want to be consistent with the rest and I would like to rename it to `excludePatterns` instead of `exclude_patterns`, because all of our other variables are following the official format. At least in the code, I have reviewed.

lukpueh:
💯for WIP/draft PRs. I'll take a closer look when the tests pass. ;) Or if you have any specific questions.

> we can still list @ayush159 as author for the code he wrote

That's very kind of you. :)

> We should also have a look on the AppVeyor pipeline.. it seems we are running into problems like in the original PR.

Could this be related to a merge conflict? Maybe try re-basing on master?

> Is this exclude_patterns in the official in-toto specification

No it's not. It is just a feature of the implementation and I don't think it needs to go into the implementation. Feel free to change the casing so that it feeds go style.

shibumi:
> Could this be related to a merge conflict? Maybe try re-basing on master?

I have no idea, when I test this branch locally via `go test ./...` everything works. Github won't show me the link to the appveyor build. `go build -o intoto ./in-toto` works as well.

shibumi:
I guess we also want to make sure to use the newest `git-go`.. they changed their remote to:

https://github.com/go-git/go-git

lukpueh:
Not sure either but there definitely seems to be a merge conflict. Would you mind trying to re-base this branch on master (resolving the conflicts) and force push. Maybe that makes Appveyor happy.

shibumi:
sure!

shibumi:
I have rebased successfuly (did a few changes on the go.mod file) and pushed a new build..

The build succeeds now, but I still want to have a look on the rest of the code.

shibumi:
I still want to change the gopkg.in remote for git-go to the github one..

shibumi:
I've updated the go-git remote + version.. but weird that github.com/go-git/go-git pulls in src-d/go-git/v4 as indirect dependency. But we are definitely using v5 in our code now :)

lukpueh:
Let me know when you want me to take a look at this.

in_toto/runlib_test.go
```diff
@@ -116,9 +144,10 @@ func TestInTotoRun(t *testing.T) {
 	linkName := "Name"
 	parameters := []map[string][]string{
 		{
-			"materialPaths": {"demo.layout.template"},
-			"productPaths":  {"foo.tar.gz"},
-			"cmdArgs":       {"sh", "-c", "printf out; printf err >&2"},
+			"materialPaths":   {"demo.layout.template"},
+			"productPaths":    {"foo.tar.gz"},
+			"cmdArgs":         {"sh", "-c", "printf out; printf err >&2"},
+			"excludePatterns": nil,
```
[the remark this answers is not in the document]
shibumi: I had a look on this via `go test -cover`, this should be covered.

in_toto/runlib_test.go
```diff
@@ -27,12 +27,40 @@ func TestRecordArtifact(t *testing.T) {
...
+	paths := []string{"foo.tar.gz", "demo.layout.template"}
+	excludePatterns := []string{"demo.layout.template"}
+	result, err := ApplyExcludePatterns(paths, excludePatterns)
+	expected := []string{"foo.tar.gz"}
+	if !reflect.DeepEqual(result, expected) {
+		t.Errorf("RecordArtifacts returned '(%s, %s)', expected '(%s, nil)'",
+			result, err, expected)
+		// os.RemoveAll("tmpdir")
+	}
+	// os.RemoveAll("tmpdir")
```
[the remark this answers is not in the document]
shibumi: I removed the lines, because I think they are not necessary anymore.

shibumi:
@lukpueh do you want to have a look on this one? I have successfully rebased on our new master, fixed a few things like formatting and added a new test function :)

shibumi:
We should make sure to close the other PR and the issue as well when we merge this PR.

shibumi:
Thanks to @ayush159 for his work :)

shibumi:
@lukpueh did you have some time to look over this? I would like to get the last two PRs merged, before I start working on the in-toto run implementation. I am afraid of too diverging codes if I just start with in-toto-run now. Would make merging unnecessarily complicated.

in_toto/runlib.go
```diff
@@ -48,6 +49,35 @@ func RecordArtifact(path string) (map[string]interface{}, error) {
...
+		temp := ParsePattern(element, nil)
+		ps = append(ps, temp)
+	}
+	m := NewMatcher(ps)
+	for _, path := range paths {
+		fileInfo, err := os.Stat(path)
+		if err != nil {
+			return nil, err
+		}
+		match := m.Match([]string{path}, fileInfo.IsDir())
```
lukpueh, line 73: Why do we care if the path is a dir?

in_toto/runlib.go
```diff
@@ -65,9 +95,19 @@ the following format:
...
+NOTE on exclude patterns:
+      - Uses gitignore from go-git library to compile gitignore-style patterns
+      - If exclude patterns are given then records artifact only for path
+        which do not match exclude patterns.
 */
-func RecordArtifacts(paths []string) (map[string]interface{}, error) {
+
+func RecordArtifacts(paths []string, excludePatterns []string) (map[string]interface{}, error) {
 	artifacts := make(map[string]interface{})
+	paths, err := ApplyExcludePatterns(paths, excludePatterns)
```
lukpueh, line 107: I fear this won't be enough. We also need to apply the patterns on the paths resolved in `filepath.Walk` below.

The `paths` slice might only contain directories that are not matched by a pattern, whereas the files in those directories resolved by `Walk` could match.

You can take a look at the reference implementation, to see how it's done there.

in_toto/runlib.go
```diff
@@ -48,6 +49,35 @@ func RecordArtifact(path string) (map[string]interface{}, error) {
...
+	// parses the exclude patterns into pattern type object
+	var ps []Pattern
+	var included []string
+	for _, element := range excludePatterns {
+		temp := ParsePattern(element, nil)
+		ps = append(ps, temp)
+	}
+	m := NewMatcher(ps)
+	for _, path := range paths {
+		fileInfo, err := os.Stat(path)
```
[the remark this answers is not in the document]
shibumi: No, because we are dealing with a gitignore matcher here. This matcher needs to know if we are dealing with a directory or not: https://pkg.go.dev/gopkg.in/src-d/go-git.v4/plumbing/format/gitignore?tab=doc#Matcher

That's why we need FileInfo to retrieve information about the actual state of the file (directory or not).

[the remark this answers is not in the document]
lukpueh: Thanks for the clarification! This also answers my question below. Would you mind adding a comment?

in_toto/runlib.go
shibumi: This is just how the gitignore matcher works: https://pkg.go.dev/gopkg.in/src-d/go-git.v4/plumbing/format/gitignore?tab=doc#Matcher

With a normal regexp matcher we would not need `fileInfo.IsDir()`: https://pkg.go.dev/gopkg.in/src-d/go-git.v4/plumbing/format/gitignore?tab=doc#Matcher

I guess the former dev chose the gitignore matcher, because it implements most of the filesystem path matching already (different platforms: Linux, Windows, Mac, easier path matching, etc)

in_toto/runlib.go
shibumi: Do you have a specific test case in mind which I could add? I will have a look on the python implementation. Would be a shame if we say we have regex support, but actually it doesn't work well :D

in_toto/runlib.go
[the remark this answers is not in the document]
shibumi: yep

in_toto/runlib.go
shibumi: I think you are right. Wouldn't it make more sense if we would apply the patterns to each string instead? So instead of returning a []string slice we should just check if this path is excluded and if so we should return a true and just move to the next path? This would need some rework of the `ApplyExcludePatterns` func

shibumi:
> May I suggest you take a look at Table Driven Tests? They seem common in go and we already do this in a bunch of places in this repo. It is a great way to remove redundant test code and to easily add new scenarios.

Sorry, just read this comment now. Table driven tests looks good to me, I think you are talking about something like this:

```
package main

import "testing"

func TestFaculty(t *testing.T) {
	tables := map[uint]uint{
		6: 720,
		2: 2,
		1: 1,
		9: 362880,
	}
	for k, v := range tables {
		result := faculty(k)
		if v != result {
			t.Errorf("Faculty %d! was incorrect, got: %d, want: %d", k, result, v)
		}
	}

}```

Right?

**Comment by lukpueh** [comment]
I just thought that [runlib_test.go#L71-L137](https://github.com/in-toto/in-toto-golang/blob/b08a29f19ac2d80dd049c5a45045dd9b0d588e00/in_toto/runlib_test.go#L71-L137)  could easily be written in table form, since each test scenario calls `RecordArtifacts` and then `DeepEqual` and only the parameters differ. But yeah, re-writing the tests is not a big priority.

Btw. I recommend watching [this talk about testing in go](https://youtu.be/8hQG7QlcLBk?t=455). I learned a thing or two from it. :)

File: in_toto/runlib.go (hunk 4 above)
Comment (lukpueh) [reply in_toto/runlib.go:107]: > Do you have a specific test case in mind which I could add?

One test case I have in mind is a file on disk `"/foo/bar"`, an exclude pattern `"bar"` (which should exclude that file), and a path `"/foo"` in the `paths` slice passed to `RecordArtifacts` and walking over files like `/foo/bar`. 

Btw. I just saw that I described the desired behavior in the commit message of https://github.com/in-toto/in-toto/commit/decf211a492b3da3cf2635410bba4f04559bfd17.

I suggest you also take a look at the tests in the reference implementation (see `tests/test_runlib.py`). and maybe coordinate with @itsbenwu, who is working on a related issue (https://github.com/in-toto/in-toto/issues/355).

> Wouldn't it make more sense if we would apply the patterns to each string instead?

Maybe. I think when when I first wrote this in https://github.com/in-toto/in-toto/commit/3ee8b67ea813830c4e228e82b6d85a10174c0752 I saw that the function I used, i.e. `fnmatch.filter` took multiple names as first argument and multiple names was what I had. :) Performance-wise it's many lightweight calls to the filter function vs. one probably more expensive call. Not sure which one performs better. Feel free to profile. To me either approach seems fine.

**Comment by shibumi** [comment]
@lukpueh ah ok, this is what you mean. This shouldn't be much work, I will refactor this section on the weekend. Should be just 10min. The talk looks great! Thanks, I will give it a shot, too.

**Comment by shibumi** [comment]
@lukpueh I've refactored the tests for the pattern matching in test tables. The code is much more readable now and it should take us less hassle in the future to add another test if we need to.

**Comment by shibumi** [comment]
The latest test case shows indeed that a path like `/$test_dir/foo.tar.gz` will not get excluded if we provide only `foo.tar.gz` as exclude pattern

**Comment by shibumi** [comment]
I have reported the gitignore behavior in the go-git repository: https://github.com/go-git/go-git/issues/108

and I am really disappointed, because there is really no magic behind this module and it's actually just calling filepath.Match:

https://golang.org/pkg/path/filepath/#Match

There seems to be no logic in the module for matching on paths via a simple filename pattern: https://github.com/go-git/go-git/blob/master/plumbing/format/gitignore/pattern.go#L92

Also even a pattern like `**/foobar.tar.gz` doesn't seem to work, although it should. I am not sure how we want to proceed in this case. We have multiple options:

1. We could apply `filepath.Base` to the path and then apply the patterns on the last element: https://golang.org/pkg/path/filepath/#Base  Problem: This would not satisfy patterns like `src/pattern.go` in paths like `/tmp/src/pattern.go`.

2. We could try to fix the behavior upstream in go-git. This would freeze this PR for now, because we need to wait for the devs.

3. We could write our own regex module

4. We could try replacing the go-git module with one of these: https://github.com/monochromegane/go-gitignore , https://github.com/denormal/go-gitignore (the latter seems very old..)

**Comment by lukpueh** [comment]
Thanks for reporting this upstream, @shibumi and for providing different viable ways to solve this. I lean towards 2 or 4 (with the more active project). 1. Seems a bit hackish and 3. overkill. Note that we don't have to use .gitignore-style matching. The exclude pattern option is not a covered by the spec, it's just a convenient implementation feature.

**Comment by shibumi** [comment]
@lukpueh I had deeper look on 4 yesterday and the more active go-gitignore project states in their readme, that they don't support the `**` operator (and this should mean that our path matching isn't supported, too).

But I am confused that the go-git gitignore doesn't support `**`. They state, they do.

**Comment by lukpueh** [comment]
@shibumi, would you kindly remove the [corresponding todo item in README](https://github.com/in-toto/in-toto-golang/blame/master/README.md#L40) with this PR?

**Comment by shibumi** [comment]
@lukpueh sure, will have a look on it.

**Comment by shibumi** [comment]
@lukpueh done, btw

**Comment by shibumi** [comment]
@lukpueh I reset this PR to current master and placed my new draft for excludePatterns on top of it. The new pathspec library should solve most of our problems, that got addressed in the comments here. What we need to do now is providing test cases and test the new functionality as good as possible.

**Comment by shibumi** [comment]
@lukpueh We should have a look at this PR next. I would like to wait with the Go 1.15 PR, because other go libs (tekton, kubesec etc are maybe not yet ready for Go 1.15)

**Comment by shibumi** [comment]
ahh great .. the joy of Windows as only operating system using the backslash as path separator. I will fix this in the pathspec library, then this should be fixed here, as well.

@lukpueh feel free to take a look already. https://github.com/in-toto/in-toto-golang/pull/53/commits/b1cda26262e8084e5a9f5ef48d976f0356a8e057 removes my old approach via ApplyGitIgnore function, because we need to check each path we are visiting via filewalk (I think you mentioned this once, but I totally forgot about this).

The new approach just calls `pathspec.GitIgnore` now directly. This works already in Linux and I test for full paths, "**" operator and "filename" match right now.

I think I will just fix the path separator in the pathspec function via checking if we are on Windows and then translating all "\" to "/", but I don't know yet if this will work.

**Comment by shibumi** [comment]
@lukpueh  as promised.. I fixed the issue in upstream go-pathspec.. it works on windows now, too :)

**Comment by SantiagoTorres** [comment]
LGTM!

File: go.mod [hunk 6]
```diff
@@ -1,5 +1,8 @@
-module github.com/in-toto/in-toto-golang
+module github.com/shibumi/in-toto-golang
```
shibumi, line 1: @SantiagoTorres  oops.. we might want to fix this. No idea how that change got in my commit :D 
I will compile a PR for this tomorrow (too late in Germany already).

Same for the go version

PR #55, opened 2020-06-05
implement symlink functionality with filepath.EvalSymlinks()

shibumi:

**Fixes issue #**:
https://github.com/in-toto/in-toto-golang/issues/32
as well as PR https://github.com/in-toto/in-toto-golang/pull/37

We are using filepath.EvalSymlinks() here, because it will
give us the last element in a symlink chain.
Therefore we don't need to worry about stack size
or recursion cycles.

requirements**:

### DIFF
--- a/in_toto/runlib.go
+++ b/in_toto/runlib.go
@@ -10,6 +10,11 @@ import (
 )
 
 /*
+ErrSymCycle signals a detected symlink cycle in our RecordArtifacts() function.
+*/
+var ErrSymCycle = errors.New("symlink cycle detected")
+
+/*
 RecordArtifact reads and hashes the contents of the file at the passed path
 using sha256 and returns a map in the following format:
 
@@ -22,7 +27,6 @@ using sha256 and returns a map in the following format:
 If reading the file fails, the first return value is nil and the second return
 value is the error.
 */
-
 func RecordArtifact(path string) (map[string]interface{}, error) {
 
 	hashObjectMap := createMap()
@@ -48,11 +52,6 @@ func RecordArtifact(path string) (map[string]interface{}, error) {
 	// Return it in a format that is conformant with link metadata artifacts
 	return hashedContentsMap, nil
 }
-
-/*
-ErrSymCycle signals a detected symlink cycle in our RecordArtifacts() function.
-*/
-var ErrSymCycle error = errors.New("symlink cycle detected")
 
 /*
 RecordArtifacts walks through the passed slice of paths, traversing
--- a/in_toto/runlib.go
+++ b/in_toto/runlib.go
@@ -56,7 +56,7 @@ var ErrSymCycle error = errors.New("symlink cycle detected")
 
 /*
 RecordArtifacts walks through the passed slice of paths, traversing
-subdirectories, and calls RecordArtifact for each file.  It returns a map in
+subdirectories, and calls RecordArtifact for each file. It returns a map in
 the following format:
 
 	{
@@ -71,6 +71,9 @@ the following format:
 
 If recording an artifact fails the first return value is nil and the second
 return value is the error.
+
+depth refers to the recursion depth. This function should be always called
+with depth = 0.
 */
 func RecordArtifacts(paths []string, depth uint) (map[string]interface{}, error) {
 	if depth > 10 {
@@ -81,7 +84,8 @@ func RecordArtifacts(paths []string, depth uint) (map[string]interface{}, error)
 	for _, path := range paths {
 		err := filepath.Walk(path,
 			func(path string, info os.FileInfo, err error) error {
-				// Abort if Walk function has a problem, e.g. path does not exist)
+				// Abort if Walk function has a problem,
+				// e.g. path does not exist
 				if err != nil {
 					return err
 				}
@@ -90,18 +94,19 @@ func RecordArtifacts(paths []string, depth uint) (map[string]interface{}, error)
 					return nil
 				}
 
-				// check for symlink and evaluate the last element in a symlink chain
-				// via filepath.EvalSymlinks. We use filepath.EvalSymlinks here, because
-				// filepath.EvalSymlinks returns last element in a symlink chain.
-				// With os.Readlink() we would just read the next element in a possible symlink chain.
+				// check for symlink and evaluate the last element in a symlink
+				// chain via filepath.EvalSymlinks. We use EvalSymlinks here,
+				// because with os.Readlink() we would just read the next
+				// element in a possible symlink chain. This would mean more
+				// iterations. infoMode()&os.ModeSymlink uses the file file
+				// type bitmap to check for a symlink.
 				if info.Mode()&os.ModeSymlink == os.ModeSymlink {
 					evalSym, err := filepath.EvalSymlinks(path)
 					if err != nil {
 						return err
 					}
-					// We recursively call RecordArtifacts() to follow the new path
-					// Note: filepath.EvalSymlinks() handles symlink chains and throws an error if
-					// it detects such a chain like: chain1 -> chain2, chain2 -> chain1
+					// We recursively call RecordArtifacts() to follow
+					// the new path.
 					evalArtifacts, evalErr := RecordArtifacts([]string{evalSym}, depth+1)
 					if evalErr != nil {
 						return evalErr
@@ -112,7 +117,8 @@ func RecordArtifacts(paths []string, depth uint) (map[string]interface{}, error)
 					return nil
 				}
 				artifact, err := RecordArtifact(path)
-				// Abort if artifact can't be recorded, e.g. due to file permissions
+				// Abort if artifact can't be recorded, e.g.
+				// due to file permissions
 				if err != nil {
 					return err
 				}
--- a/in_toto/runlib.go
+++ b/in_toto/runlib.go
@@ -80,6 +80,30 @@ func RecordArtifacts(paths []string) (map[string]interface{}, error) {
 				if info.IsDir() {
 					return nil
 				}
+
+				// check for symlink and evaluate the last element in a symlink chain
+				// via filepath.EvalSymlinks. We use filepath.EvalSymlinks here, because
+				// filepath.EvalSymlinks returns last element in a symlink chain.
+				// With os.Readlink() we would just read the next element in a possible symlink chain.
+				if info.Mode()&os.ModeSymlink == os.ModeSymlink {
+					evalSym, err := filepath.EvalSymlinks(path)
+					if err != nil {
+						return err
+					}
+					// We recursively call RecordArtifacts() to follow the new path
+					// Note: filepath.EvalSymlinks() handles symlink chains and throws an error if
+					// it detects such a chain like: chain1 -> chain2, chain2 -> chain1
+					evalArtifacts, evalErr := RecordArtifacts([]string{evalSym})
+					if evalErr != nil {
+						return evalErr
+					}
+					for key, value := range evalArtifacts {
+						artifacts[key] = value
+					}
+					return nil
+
+				}
+
 				artifact, err := RecordArtifact(path)
 				// Abort if artifact can't be recorded, e.g. due to file permissions
 				if err != nil {
--- a/in_toto/runlib.go
+++ b/in_toto/runlib.go
@@ -76,14 +76,9 @@ If recording an artifact fails the first return value is nil and the second
 return value is the error.
 */
 func RecordArtifacts(paths []string) (evalArtifacts map[string]interface{}, err error) {
-	// check if visitedSymlinks has been initialized
-	// if not: generate a new set
-	if visitedSymlinks == nil {
-		visitedSymlinks = NewSet()
-	}
+	// Make sure to initialize a fresh hashset for every RecordArtifacts call
+	visitedSymlinks = NewSet()
 	evalArtifacts, err = recordArtifacts(paths)
-	// tear down visitedSymlinks
-	visitedSymlinks = nil
 	// pass result and error through
 	return evalArtifacts, err
 }
--- a/in_toto/runlib.go
+++ b/in_toto/runlib.go
@@ -9,10 +9,11 @@ import (
 	"syscall"
 )
 
-/*
-ErrSymCycle signals a detected symlink cycle in our RecordArtifacts() function.
-*/
+// ErrSymCycle signals a detected symlink cycle in our RecordArtifacts() function.
 var ErrSymCycle = errors.New("symlink cycle detected")
+
+// visitedSymlinks is a hashset that contains all paths that we have visited.
+var visitedSymlinks map[string]struct{}
 
 /*
 RecordArtifact reads and hashes the contents of the file at the passed path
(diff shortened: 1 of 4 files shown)


shibumi:
mh ok this is weird.. when I run these tests locally they work...:
```
=== RUN   TestSymlinkToFile
--- PASS: TestSymlinkToFile (0.00s)
=== RUN   TestSymlinkToFolder
--- PASS: TestSymlinkToFolder (0.00s)
```

is this again some Windows related stuff?

shibumi:
mhh looks like an error in path translation between Linux paths and Windows paths?!

```
json: unsupported type: func(*testing.T)
open bad/path: The system cannot find the path specified.
--- FAIL: TestSymlinkToFolder (0.00s)
    runlib_test.go:82: RecordArtifacts returned '(map[symTest\symTest2\symTmpfile:map[sha256:ba7816bf8f01cfea414140de5dae2223b00361a396177a9cb410ff61f20015ad]], %!s(<nil>))', expected '(map[symTest/symTest2/symTmpfile:map[sha256:ba7816bf8f01cfea414140de5dae2223b00361a396177a9cb410ff61f20015ad]], nil)'
```

I have also no idea where this `"%!s(<nil>)"` comes from

shibumi:
Doesn't windows support SymLinks?

shibumi:
I think I fixed the file path matching via using: `filepath.FromSlash()`.. I am just wondering why this works for all the other tests.. lol.

shibumi:
Mh I think I will add 1-2 more tests to the symlink code. I am not convinced yet, that it's secure without a max recursion parameter

-> draft

shibumi:
I've added a test for this PR.. seems I were wrong and filepath.EvalSym() can not detect symlinks. :S 
The newest test proofs this.

I could re-implement the recursion counter from the other PR or create a history or something.

shibumi:
@lukpueh Can you have a look on this PR, too? I have reimplemented most of the PR #37.
The only difference here is that we return an error if detect a symlink cycle instead of going further.
I really think there should be no valid reason why a software project or any other project should build cycles with symlinks

shibumi:
@lukpueh I have also added our first custom error. The error handling got some nice changes in go 1.13, we can use these changes and return our own library specific errors instead. This makes developers life easier, because they don't need to compare strings anymore.
They can just do:

```go
if err == errors.Is(err, OurCustomErr){ ..}
```

shibumi:
Shall I squash this PR into one commit? would make it easier, if we encounter problems to revert it.

in_toto/runlib.go
```diff
@@ -80,6 +89,28 @@ func RecordArtifacts(paths []string) (map[string]interface{}, error) {
...
+				// filepath.EvalSymlinks returns last element in a symlink chain.
+				// With os.Readlink() we would just read the next element in a possible symlink chain.
+				if info.Mode()&os.ModeSymlink == os.ModeSymlink {
+					evalSym, err := filepath.EvalSymlinks(path)
+					if err != nil {
+						return err
+					}
+					// We recursively call RecordArtifacts() to follow the new path
+					// Note: filepath.EvalSymlinks() handles symlink chains and throws an error if
+					// it detects such a chain like: chain1 -> chain2, chain2 -> chain1
```
lukpueh, line 104: Is this comment still true? How does this it relate to your discovery in https://github.com/in-toto/in-toto-golang/pull/55#issuecomment-641152088?

in_toto/runlib.go
```diff
@@ -66,7 +72,10 @@ the following format:
 If recording an artifact fails the first return value is nil and the second
 return value is the error.
 */
-func RecordArtifacts(paths []string) (map[string]interface{}, error) {
+func RecordArtifacts(paths []string, depth uint) (map[string]interface{}, error) {
+	if depth > 10 {
```
lukpueh, line 76: Why 10?

in_toto/runlib_test.go
```diff
@@ -27,6 +28,117 @@ func TestRecordArtifact(t *testing.T) {
...
+		t.Errorf("Could not create tmpdir: %s", err)
+	}
+
+	// we need to get the current working directory here, otherwise
+	// os.Symlink() will create a wrong symlink
+	dir, err := os.Getwd()
+	if err != nil {
+		t.Error(err)
+	}
+	// create a cycle ./symlinkCycle/symCycle.sym -> ./symlinkCycle/
```
lukpueh, line 122: Would you mind adding a test for an indirect circle? E.g.
/bar/foo.sym --> /foo/ 
/foo/bar.sym --> /bar/

lukpueh:
Many thanks for the great work, @shibumi. I left a few minor suggestions and questions.

in_toto/runlib.go
```diff
@@ -66,7 +72,10 @@ the following format:
 If recording an artifact fails the first return value is nil and the second
 return value is the error.
 */
-func RecordArtifacts(paths []string) (map[string]interface{}, error) {
+func RecordArtifacts(paths []string, depth uint) (map[string]interface{}, error) {
```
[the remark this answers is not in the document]
shibumi: Oh yes, this makes sense. Totally forgot about this, thanks.

in_toto/runlib.go
```diff
@@ -80,6 +89,28 @@ func RecordArtifacts(paths []string) (map[string]interface{}, error) {
 				if info.IsDir() {
 					return nil
 				}
+
+				// check for symlink and evaluate the last element in a symlink chain
+				// via filepath.EvalSymlinks. We use filepath.EvalSymlinks here, because
+				// filepath.EvalSymlinks returns last element in a symlink chain.
+				// With os.Readlink() we would just read the next element in a possible symlink chain.
```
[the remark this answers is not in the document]
shibumi: Yep, will do

in_toto/runlib.go
```diff
@@ -80,6 +89,28 @@ func RecordArtifacts(paths []string) (map[string]interface{}, error) {
...
+				if info.Mode()&os.ModeSymlink == os.ModeSymlink {
```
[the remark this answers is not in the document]
shibumi: It's strange that Golang provides a `filemode.isDir()` function, but no `filemode.isSymlink()` function. This is something I could try to add to Golang in the future ^^

For now I will add a comment + shorten the comment about filepath.EvalSymlinks

in_toto/runlib.go
shibumi: This will get updated, too

in_toto/runlib.go
shibumi: 10 is just an artificial number I have inherit from the former PR. We should actually think about a better number here or at least a good description, why we choose 10 and not 9 or 11 or 9000.

in_toto/runlib_test.go
shibumi: This makes sense

in_toto/runlib.go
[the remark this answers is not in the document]
shibumi: This should be solved.. I just wonder if we can set a default value for parameters like in Python:
`def RecordArtifacts(paths, depth=0):`

I will google a bit and see if this is manageable, because otherwise I am afraid that always calling this function with depth=0 is very error prone. People really need to make sure to read the docs.

in_toto/runlib.go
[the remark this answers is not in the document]
shibumi: This should be fixed.. I leave resolving the conversation to you, just in case you found something else or I misunderstood you.

in_toto/runlib.go
```diff
@@ -48,6 +49,11 @@ func RecordArtifact(path string) (map[string]interface{}, error) {
 	return hashedContentsMap, nil
 }
 
+/*
+SymCycleErr signals a detected symlink cycle in our RecordArtifacts() function.
+*/
+var SymCycleErr error = errors.New("symlink cycle detected")
```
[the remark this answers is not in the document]
shibumi: This should be fixed.. I leave resolving the conversation to you, just in case you found something else or I misunderstood you.

in_toto/runlib.go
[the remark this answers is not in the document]
shibumi: This should be fixed.. I leave resolving the conversation to you, just in case you found something else or I misunderstood you.

in_toto/runlib.go
shibumi: This should be fixed.. I leave resolving the conversation to you, just in case you found something else or I misunderstood you.

in_toto/runlib.go
shibumi: We could set the go runtime stack size here. Problem: It's different for the architecture. On Linux 64 bit it's usually 1GB.

But I have maybe a better solution, but this would need some more work for this PR.
What about using a history of visited symlinks? We could just add a visited symlink to a slice or hashset (`map[string]struct{}` for lookup time of O(1)). If we encounter a symlink twice, we just ignore it. What do you think about this, @lukpueh?

in_toto/runlib_test.go
shibumi: I've added another test for your indirect cycle.

in_toto/runlib.go
shibumi: I had to try this. We are now using a symlink history via a global hashset variable. This has better performance, because we don't need to cycle n times, before detecting a symlink.
Instead of we are checking if we have seen this symlink already, if so, we will break the recursion and error our Symlink Cycle error.

Feel free to check it out.

in_toto/runlib.go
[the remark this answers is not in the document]
lukpueh: I fear go does not have that feature. But I wouldn't worry too much about usage mistakes. If you pass a parameter to a function, you usually want to know what it is for, right? Also, although we export this function as public API, its mostly for internal use by `InTotoRun`.

in_toto/runlib.go
lukpueh: I was hoping you would implement something like this, because I think the depth parameter would have failed detecting some indirect circles. Thanks for putting more efforts into this, @shibumi! 🥇

in_toto/runlib.go
```diff
@@ -67,21 +73,61 @@ If recording an artifact fails the first return value is nil and the second
...
 				// Don't hash directories
 				if info.IsDir() {
 					return nil
 				}
+
+				// check for symlink and evaluate the last element in a symlink
+				// chain via filepath.EvalSymlinks. We use EvalSymlinks here,
+				// because with os.Readlink() we would just read the next
+				// element in a possible symlink chain. This would mean more
+				// iterations. infoMode()&os.ModeSymlink uses the file file
```
lukpueh, line 100:
```suggestion
				// iterations. infoMode()&os.ModeSymlink uses the file
```

in_toto/runlib.go
```diff
@@ -67,21 +73,61 @@ If recording an artifact fails the first return value is nil and the second
...
+				// type bitmap to check for a symlink.
```
lukpueh, line 101:
```suggestion
				// type bitmask to check for a symlink.
```

in_toto/runlib.go
```diff
@@ -1,13 +1,20 @@
...
 	"os/exec"
 	"path/filepath"
 	"syscall"
 )
 
+// ErrSymCycle signals a detected symlink cycle in our RecordArtifacts() function.
+var ErrSymCycle = errors.New("symlink cycle detected")
+
+// visitedSymlinks is a hashset that contains all paths that we have visited.
+var visitedSymlinks map[string]struct{}
```
lukpueh, line 16: FYI, we have a nice custom `Set` implementation with `Add` and `Has` methods in `util.go`. :) I suggest you use that.

shibumi:
I will squash the last 2 nonsense commits from github UI to one 'refactor comments' commit. Shouldn't use github for applying changes directly .. sigh

in_toto/runlib.go
shibumi: Ah cool, yes this makes sense. Would increase the readability a lot.

shibumi:
> Thanks for coming up with and implementing this new approach, @shibumi! It looks a lot better than what we had before. Will this also work if we called `RecordArtifacts` twice? I think you need to reset `visitedSymlinks` at some point.

We could pass a set as argument, that way we could control the passed set and on recursion depth 0 devs would need to pass an empty set as initialization. It's the only solution I come up with, but I will think about this till Friday afternoon.

lukpueh:
💯  Thanks, that's what I usually do. I like the feature from a reviewer's perspective, but I also think there should be no  "Update XYZ" commits.

lukpueh:
> We could pass a set as argument, that way we could control the passed set and on recursion depth 0 devs would need to pass an empty set as initialization.

Sounds like a good idea. Or we add a wrapper like so:
```
# pseudo-code
function RecordArtifacts(paths):
  setup global var
  _recordArtifacts(paths) # recurse but don't visit symlinks more than once
  tear down global var
```

shibumi:
I like the wrapper variant. This is better than passing a map. We could do it like the other big Golang libraries and just write `recordArtifacts`, for example os.Rename(): https://golang.org/src/os/file.go?s=9921:9963#L313

in_toto/runlib.go
shibumi: I've implemented a wrapper around recordArtifacts + use the Set from utils.

shibumi:
@lukpueh I am not sure if `visitedSymlinks = nil` is enough for 'removing' the variable and tearing it down correctly. This should work, if we want to be sure I could also implement a `RemoveAll` function for our Set and call that one. 5min of Google said setting it nil should be enough.

in_toto/runlib.go
```diff
@@ -66,22 +75,88 @@ the following format:
 If recording an artifact fails the first return value is nil and the second
 return value is the error.
 */
-func RecordArtifacts(paths []string) (map[string]interface{}, error) {
+func RecordArtifacts(paths []string) (evalArtifacts map[string]interface{}, err error) {
+	// check if visitedSymlinks has been initialized
+	// if not: generate a new set
+	if visitedSymlinks == nil {
+		visitedSymlinks = NewSet()
+	}
```
lukpueh, line 83: Do we want to support a use-case where the visitedSymlinks Set is initialized outside of this function? I feel like (re-)initializing it no matter what, or raising an error if it is already initialized come both more expected.

shibumi: We could allow passing the set to the function and if this set is nil we can initialize a global set, but do we still need a global variable and a wrapper if we allow this? And If we allow passing a set to the function, shouldn't the function return a set then?

We could also work with Go receivers (I have no idea if this is supported by our set. It's just a map with empty structs as values.. It seems to be possible according to: https://stackoverflow.com/questions/49176090/map-as-a-method-receiver):
```
func (visitedSymlinks Set) RecordArtifacts(...)
```

shibumi: Ah ok.. Now I understood your second sentence, sorry. Yes it makes sense to always initialize it no matter what. Checking for nil, is wrong here, because we want to make sure that the set is always fresh initialized. I think this is an artifact from renaming the methods.. this would have made sense if we call RecordArtifacts (the wrapper) recursively, but we don't.

shibumi: I am going to fix this

shibumi:
@lukpueh I've fixed the nil check for the visitedSymlinks hashset. Ping me if you find something else :)

shibumi:
@lukpueh checks passed, ready to merge. When merged, I will use the current master as base for the intoto-run implementation.

Issue #75, opened 2020-08-25
Windows Compatibility


shibumi:
**Description of issue or feature request**:

I had a a deep-dive into Windows today via installing a Windows10 Vagrant box, starting Jetbrains GoLand in it and executing the tests for master and for my branches.

I have bad news.

Did we ever execute tests via appveyor on Windows, besides go fmt?
If so, then I am interested in what appveyor is actually doing, because the tests on Windows definitely fail.

There are several issues. First I thought it's the Github Runner, that is behaving odd, but in reality it's our code.
Somehow generateKeyID calculates a different checksum on Windows. I have no explanation for this.

Here is what I did:

1. First I compared file sizes of our test data files, these were differently. Later I read at StackOverflow, that these are just different metrics: https://askubuntu.com/questions/573586/files-sizes-differs-from-windows-to-linux#:~:text=The%20different%20file%20structure%20which,sizes%20in%20different%20operating%20systems.

2. So I decided to calculate the checksum on Linux via `md5sum` and `sha256sum` and on Windows via `CertUtil -hashfile <input file> sha256`. Well.. and all of our test data files, had a different check sum. Do you have any explanation for this?

I tried setting `.gitattributes` and `eol=lf` and such stuff, but it didn't help.

What is even more odd is, that we use `pem.Decode()`, so actually the file format is irrelevant. But still we are ending up with a different keyID. I expect an issue in `EncodeCanonicalData`, but I can't really explain what is going on there.

**Current behavior**:
Right now we are not Windows compatible and I doubt that we ever were windows compatible.

**Expected behavior**:
We should be compatible with windows.

PS:
Do we have checks for Windows/Linux compatibility with in-toto-py and/or TUF?

CC: @SantiagoTorres @trishankatdatadog maybe you have an explanation for this.

SantiagoTorres:
Hmm, first I wanted to share this admittedly not great repo to help you set up a dev [machine in windows](https://github.com/in-toto/windows-setup). I'm pretty sure I've used py-in-toto to generate metadata on windows and consume it on linux and vice versa.

> Somehow generateKeyID calculates a different checksum on Windows. I have no explanation for this.

I thought I had explained this on slack, but I think this has to do to the fact that windows defaults to utf-16, so you may be hashing a bunch of additional zeros. Have you printed the buffers themselves before they are being passed to the hashing function? (i.e., dump the whole memory of the hashable buffer and then compare them)

shibumi:
[...]

PR #76, opened 2020-08-26
Never use pemBytes directly

shibumi:
**Fixes issue #**:

https://github.com/in-toto/in-toto-golang/issues/75

We cannot use the pemBytes directly, because there might be different
line separators on different operating systems (CRLF vs LF).
Therefore we are always decoding the pemBytes now and we are only
working with decoded key bytes. These formats are operating system independent.

requirements**:

### DIFF
--- a/in_toto/keylib.go
+++ b/in_toto/keylib.go
@@ -119,19 +119,19 @@ func (k *Key) generateKeyID() error {
 }
 
 /*
-generatePublicPEMBlock creates a "PUBLIC KEY" PEM block from public key byte data.
-If successful it returns PEM block as []byte slice. This function should always
-succeed, if pubKeyBytes is empty the PEM block will have an empty byte block.
+generatePEMBlock creates a PEM block from scratch via the keyBytes and the pemType.
+If successful it returns a PEM block as []byte slice. This function should always
+succeed, if keyBytes is empty the PEM block will have an empty byte block.
 Therefore only header and footer will exist.
 */
-func generatePublicPEMBlock(pubKeyBytes []byte) []byte {
+func generatePEMBlock(keyBytes []byte, pemType string) []byte {
 	// construct PEM block
-	publicKeyPemBlock := &pem.Block{
-		Type:    "PUBLIC KEY",
+	pemBlock := &pem.Block{
+		Type:    pemType,
 		Headers: nil,
-		Bytes:   pubKeyBytes,
+		Bytes:   keyBytes,
 	}
-	return pem.EncodeToMemory(publicKeyPemBlock)
+	return pem.EncodeToMemory(pemBlock)
 }
 
 /*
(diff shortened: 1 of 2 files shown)


shibumi:
A few things:

* More comments?
* Also I have the feeling, that we are doing things twice again. With this patch, we will decode the pemBytes and later decode them again for the public Key. Quick Idea: We return two variables with decode and Parse.

shibumi:
The second commit introduces a second return param to `decodeAndParse`. This way we can work with the pemData if needed and throw it away if we don't need it.

lukpueh (commented):
Very cool stuff, @shibumi!

Two small observations:
- Do we still need to call `TrimSpace` in `setKeyComponents`, now that we re-generate the normalized PEM Block?
- Would it be worth to define constants for the `pemType` strings  passed to `generatePEMBlock` in `setKeyComponents`?

shibumi:
> Do we still need to call TrimSpace in setKeyComponents, now that we re-generate the normalized PEM Block?

need to investigate this.

> Would it be worth to define constants for the pemType strings passed to generatePEMBlock in setKeyComponents?

The benefit of it would be minor, but might makes sense. I will add a few constants for these.

shibumi:
@lukpueh 
I have added constants + I have checked if we can remove the string.Trim operation. We cannot. If I remove it, this will alter all keyIDs, I suspect it's adding a newline at the end or anything in this direction.

shibumi:
> I have added constants + I have checked if we can remove the string.Trim operation. We cannot. If I remove it, this will alter all keyIDs, I suspect it's adding a newline at the end or anything in this direction.

Ok, I have verified this with a debugger.

**with string.TrimSpace**:
```
-----BEGIN PUBLIC KEY-----
MIIBojANBgkqhkiG9w0BAQEFAAOCAY8AMIIBigKCAYEAxPX3kFs/z645x4UOC3KF
Y3V80YQtKrp6YS3qU+Jlvx/XzK53lb4sCDRU9jqBBx3We45TmFUibroMd8tQXCUS
e8gYCBUBqBmmz0dEHJYbW0tYF7IoapMIxhRYn76YqNdl1JoRTcmzIaOJ7QrHxQrS
GpivvTm6kQ9WLeApG1GLYJ3C3Wl4bnsI1bKSv55Zi45/JawHzTzYUAIXX9qCd3Io
HzDucz9IAj9Ookw0va/q9FjoPGrRB80IReVxLVnbo6pYJfu/O37jvEobHFa8ckHd
YxUIg8wvkIOy1O3M74lBDm6CVI0ZO25xPlDB/4nHAE1PbA3aF3lw8JGuxLDsetxm
fzgAleVt4vXLQiCrZaLf+0cM97JcT7wdHcbIvRLsij9LNP+2tWZgeZ/hIAOEdaDq
cYANPDIAxfTvbe9I0sXrCtrLer1SS7GqUmdFCdkdun8erXdNF0ls9Rp4cbYhjdf3
yMxdI/24LUOOQ71cHW3ITIDImm6I8KmrXFM2NewTARKfAgMBAAE=
-----END PUBLIC KEY-----
```

**without string.TrimSpace**:
```
-----BEGIN PUBLIC KEY-----
MIIBojANBgkqhkiG9w0BAQEFAAOCAY8AMIIBigKCAYEAxPX3kFs/z645x4UOC3KF
Y3V80YQtKrp6YS3qU+Jlvx/XzK53lb4sCDRU9jqBBx3We45TmFUibroMd8tQXCUS
e8gYCBUBqBmmz0dEHJYbW0tYF7IoapMIxhRYn76YqNdl1JoRTcmzIaOJ7QrHxQrS
GpivvTm6kQ9WLeApG1GLYJ3C3Wl4bnsI1bKSv55Zi45/JawHzTzYUAIXX9qCd3Io
HzDucz9IAj9Ookw0va/q9FjoPGrRB80IReVxLVnbo6pYJfu/O37jvEobHFa8ckHd
YxUIg8wvkIOy1O3M74lBDm6CVI0ZO25xPlDB/4nHAE1PbA3aF3lw8JGuxLDsetxm
fzgAleVt4vXLQiCrZaLf+0cM97JcT7wdHcbIvRLsij9LNP+2tWZgeZ/hIAOEdaDq
cYANPDIAxfTvbe9I0sXrCtrLer1SS7GqUmdFCdkdun8erXdNF0ls9Rp4cbYhjdf3
yMxdI/24LUOOQ71cHW3ITIDImm6I8KmrXFM2NewTARKfAgMBAAE=
-----END PUBLIC KEY-----

````

lukpueh (approved):
Thanks for checking if we can remove TrimSpace!

shibumi:
Just one addition for being complete:

I wondered if this newline is some sort of standard in PEM, but turns out, that this is the job of the parser and PEM (RFC7468) states, that there can be whitespaces in the PEM block.

See also: https://tools.ietf.org/html/rfc7468#section-2

>  Data before the encapsulation boundaries are
   permitted, and parsers MUST NOT malfunction when processing such
   data.  Furthermore, parsers SHOULD ignore whitespace and other non-
   base64 characters and MUST handle different newline conventions.

PR #77, opened 2020-08-26
Add Logo + Remove todo

shibumi:
**Fixes issue #**:
none

add logo + remove todo (I think parameter layout substitution is already supported, right?)
See also: #29 and #38 

requirements**:


shibumi:
I think the image will get scaled automatically, because READMEs have a fixed border as preview README on a github repository.

Or shall I scale it down?

lukpueh:
I wonder if we even want the logo in the README, and if so what other repos under the in-toto GitHub org should have the logo in the README. If we add it here, we should definitely also add it to the reference implementation's README.

But maybe it's enough to have it [as organization logo](https://github.com/in-toto), which btw. also seems to be used as default for social media preview for each repo under the org.

What do you think?

lukpueh:
Btw. I just realized that we can change the title and short description in the README. in-toto-golang is no longer just in-toto verification.

shibumi:
> But maybe it's enough to have it as organization logo, which btw. also seems to be used as default for social media preview for each repo under the org.

Sounds, good. We can still change the smaller issues we have found in this PR, like:

> Btw. I just realized that we can change the title and short description in the README. in-toto-golang is no longer just in-toto verification.

shibumi:
@lukpueh 

I have removed the logo + changed the description and title.

lukpueh (approved):

Issue #31, opened 2019-03-08
Support multiple hashing algorithms


SantiagoTorres:
**Description of issue or feature request**: The current implementation is fixed to sha256. While this is is enough for a MVP, we need to support plug-able and extendable hash functions.

**Current behavior**: It is not possible to generate hashes using any other algorithm than sha256.

**Expected behavior**: It should be possible to programmatically select the hashing algorithm to compute artifact integrity and for signature verification.

hisenberg08:
I have added a couple of more hash functions in the code but it is still a static list of hash algorithms. 
Also the hash functions I am using are the one provided by the crypto library in go.
Is it okay if I keep using the same library?
How exactly do we need to make it plug-able ?
One option could be to read the list of hash functions from a configuration file.

SantiagoTorres:
Hi, @hisenberg08!

 This does seem reasonable,

What I'd like to have is an adapter in which we can easily add support for more hash algos if needed. Why don't you share what we have and we start discussing a path forward?

lukpueh:
Thanks for working on this, @hisenberg08! 

You might want to take a look at the various `create.*signature` and `verify.*signature` functions in [`securesystemslib`](https://github.com/secure-systems-lab/securesystemslib) (i.e. the crypto library we use for our Python reference implementation).

Most of those functions have some sort of *signature scheme* parameter, which encodes the hashing algorithm used to create or verify a signature. IIRC they currently all support only one scheme, but to prepare the same sort of extensibility is what we want here in the golang implementation.
Same goes for hashing algorithms used to generate a keyid (see `securesystemslib.keys._get_keyid`).

Both signature `scheme` and `keyid_hash_algorithms` should be defined directly on the corresponding key (see the various schema definitions in `securesystemslib.formats`).

And there is one more place where we want to support the use of different hashing algorithms, that is when we record *materials* and *products* (see [`in_toto.runlib._hash_artifact`](https://github.com/in-toto/in-toto/blob/develop/in_toto/runlib.py#L54)).

Please reach out if you have any further questions!

SantiagoTorres:
partly fixed by #39 . A CLI/API change is needed to allow for these algorithms to be completely pluggable.

Issue #67, opened 2020-08-16
ecdsa-sha2-nistp384 support


shibumi:
**Description of issue or feature request**:
The securesystemslib supports `ecdsa-sha2-nistp384` as scheme. We should catch up with the securesystemslib and provide
support for this curve as well. I suggest to use Go's Hash interface for setting the correct hash algorithm: https://godoc.org/crypto#Hash

This could be also useful if we want to support more RSA schemes.

**Current behavior**:
Right now we only support `ecdsa-sha2-nistp256`.

**Expected behavior**:
We should catch up with the securesystemslib and support `ecdsa-sha2-nistp384`, too.

shibumi:
A similar issue I have spotted is this one here:
https://github.com/in-toto/in-toto-golang/issues/31

lukpueh:
Kindly fixed by @shibumi in #78

Issue #65, opened 2020-08-14
ecdsa curve sanity checks


shibumi:
**Description of issue or feature request**:
The current ecdsa implementation does not do sanity checks for different ecdsa curves. Right now, we trust the developer, that
the dev will pick the right curve name in the `key.Scheme` string. Theoretically and practically the dev could just load an ecdsa key with another curve, than specified in the scheme.

**Current behavior**:
We do no sanity checks for ecdsa curves, thus it's possible that ecdsa key and `key.Scheme` do not match.

**Expected behavior**:
We should validate if the curve used in the ecdsa key matches the curve specified in the `key.Scheme` variable.

lukpueh:
Kindly fixed by @shibumi in #78

PR #78, opened 2020-08-29
implement multi hash support

shibumi:
**Fixes issue #**:

This PR is intended to fix: https://github.com/in-toto/in-toto-golang/issues/31 and https://github.com/in-toto/in-toto-golang/issues/67 and https://github.com/in-toto/in-toto-golang/issues/65

It will maybe also provide different rsassa-pss schemes like: rsassa-pss-sha512

In this PR we remove our old hash implementation and use Go's hash.Hash interface instead.
The goal is to use the new hashlib for all hash related tasks. We should be able to use this in RecordArtifact
and in the keylib for generating hashes  for example for rsassa-pss-sha256 or ecdsa-sha2-nistp384.

requirements**:


shibumi:
@lukpueh @SantiagoTorres  this is just a first draft, but what do you think about it? We could introduce a new parameter for RecordArtifacts called `hashAlgorithms`. I really want to remove this static `hashFunc := []string{"sha256"}` there.

shibumi:
https://github.com/in-toto/in-toto-golang/pull/78/commits/8539b0117d3ed22069408ed658ca9152ff395fe9 implements ecdsa-sha2-nistp384 support.

What is bothering me right now is dynamic hashing in generateKeyID(). How do we do this? Why do we even have a keyIdHashAlgorithms slice? I mean, how do we save multiple keyIDs? Right now the keyID is just a string.. do we just concatenate them? Wouldn't a hashmap be better for keyIDs like: "hashing algo": "hash"?

shibumi:
So, I had a look on the securesystemslib today.. and it looks like we never use the generated keyIDs by  multiple hash algorithms:

https://github.com/secure-systems-lab/securesystemslib/blob/876772c800138c617b30b0e33cc013b1b70b9c4c/securesystemslib/interface.py#L594

Instead we are using the default "sha256" hash algorithm for the keyIDs there, too. What is the exact purpose of the keyID_Hash_Algorithms parameter then?

lukpueh:
See https://github.com/secure-systems-lab/securesystemslib/issues/251#issuecomment-683670389 for some context about hash algorithms in in-toto/securesystemslib/

shibumi:
A few observations I did today:

1. our Test key "frank" is not a 256 key, it's actually a p521 key, so it should use SHA512 according to the specification: https://tools.ietf.org/html/rfc5656#section-6.2.1

2. We should classify ecdsa keys via their curve size and not via their scheme. Nevertheless we should do a sanity check if the scheme equals the curve size.

shibumi:
I will add a few missing tests and then this PR here should be ready to review.
Sadly, it decreases our test coverage a bit, because we have a few routines, that can't be checked, like:

```go
	if _, ok := hashMapping["sha256"]; !ok {
				return ErrUnsupportedHashAlgorithm
			}
			hashed = hashToHex(hashMapping["sha256"](), unverified)
```

We actually never run into the error here, because our hashMapping is hardcoded. Of course it will always succeed.
But I don't want to remove the check, because we may play around with the supported hashes in the future and without the check we would run into a panic. Opinions?

shibumi:
@lukpueh feel free to review. The only thing what I may think about is to build a function for these blocks here:
```go
if err := matchEcdsaScheme(curveSize, key.Scheme); err != nil {
    return ErrCurveSizeSchemeMismatch
}
if _, ok := hashMapping["sha512"]; !ok {
    return ErrUnsupportedHashAlgorithm
}
hashed = hashToHex(hashMapping["sha512"](), unverified)
```

but what would be a good name for it? "generateEcdsaHash"?

shibumi:
mh I don't like that the conflict got resolved via merging.. shall I rebase instead or is this fine @lukpueh ?

shibumi:
@lukpueh checked out newest master, rebased on master

in_toto/keylib.go
```diff
@@ -390,9 +389,12 @@ func GenerateSignature(signable []byte, key Key) (Signature, error) {
 		}
 		switch key.Scheme {
 		case rsassapsssha256Scheme:
-			hashed := sha256.Sum256(signable)
+			if _, ok := hashMapping["sha256"]; !ok {
+				return Signature{}, ErrUnsupportedHashAlgorithm
+			}
```
lukpueh, line 394: We know that `"sha256"` is in `hashMapping`, right? I think we should trust our own code.

in_toto/hashlib.go
```diff
@@ -0,0 +1,30 @@
...
+
+/*
+getHashMapping returns a mapping from hash algorithm to supported hash
+algorithm.
+*/
+func getHashMapping() map[string]func() hash.Hash {
+	return map[string]func() hash.Hash{
+		"sha256": sha256.New,
+		"sha512": sha512.New,
+		"sha384": sha512.New384,
```
lukpueh, line 17: Do you think it's worth defining string constants for the algorithm names, given that we use them elsewhere in the code too.

in_toto/keylib.go
lukpueh: Same comment applies to the rest of the diff where you check if an expected value is actually in `hashMapping`.

in_toto/keylib.go
```diff
@@ -411,29 +413,54 @@ func GenerateSignature(signable []byte, key Key) (Signature, error) {
...
-			if err != nil {
-				return signature, nil
+		curveSize := parsedKey.(*ecdsa.PrivateKey).Curve.Params().BitSize
+		var hashed []byte
+		// implement https://tools.ietf.org/html/rfc5656#section-6.2.1
+		// We determine the curve size and choose the correct hashing
+		// method based on the curveSize
+		switch {
+		case curveSize <= 256:
+			if err := matchEcdsaScheme(curveSize, key.Scheme); err != nil {
```
lukpueh, line 423: 💯 for matching scheme with curve size! But why not do it only once before the switch?

lukpueh: Same goes for use of `matchEcdsaScheme` in `VerifySignature`.

in_toto/keylib.go
```diff
@@ -517,23 +548,48 @@ func VerifySignature(key Key, sig Signature, unverified []byte) error {
...
-			if err := ecdsa.Verify(parsedKey.(*ecdsa.PublicKey), hashed[:], ecdsaSignature.R, ecdsaSignature.S); err == false {
-				return ErrInvalidSignature
+			if _, ok := hashMapping["sha512"]; !ok {
+				return ErrUnsupportedHashAlgorithm
 			}
+			hashed = hashToHex(hashMapping["sha512"](), unverified)
 		default:
-			// supported key schemes will get checked in validateKey
-			panic("unexpected Error in GenerateSignature function")
+			panic("unexpected Error in VerifySignature function")
```
lukpueh, line 582: Oh, that must have slipped through in https://github.com/in-toto/in-toto-golang/pull/56. Thanks for fixing!

in_toto/keylib.go
```diff
@@ -497,14 +525,17 @@ func VerifySignature(key Key, sig Signature, unverified []byte) error {
 		}
 		switch key.Scheme {
 		case rsassapsssha256Scheme:
-			hashed := sha256.Sum256(unverified)
+			if _, ok := hashMapping["sha256"]; !ok {
+				return ErrUnsupportedHashAlgorithm
+			}
+			hashed := hashToHex(hashMapping["sha256"](), unverified)
```
lukpueh, line 531: No that `hashed` is no longer an array you don't have to convert it to a slice below in `rsa.VerifyPSS` and `ecdsa.Verify`, i.e. you can do `s/hashed[:]/hashed`.

lukpueh (changes requested):
🎉 Very cool stuff as usual, @shibumi! I haven't looked at the tests in detail yet, but the rest looks pretty good to me, modulo some minor nits that I pointed out inline. Please address and I will review the tests in the meanwhile.

lukpueh (changes requested):
Tests look quite good too.

One thing I noticed is that some of the docstrings need to be updated to account for your recent changes, in particular:
- RecordArtifact, RecordArtifacts, recordArtifacts (new parameter)
- LoadKey (new schemes)
- GenerateSignature, VerifySignature (new errors)

It might be worth to double check other functions you changed as well.

in_toto/keylib_test.go
```diff
@@ -87,8 +93,8 @@ func TestLoadKeyErrors(t *testing.T) {
 		err            error
 	}{
 		{"not existing file", "inToToRocks", "rsassa-pss-sha256", []string{"sha256", "sha512"}, os.ErrNotExist},
-		{"existing, but invalid file", "demo.layout.template", "ecdsa-sha2-nistp256", []string{"sha512"}, ErrNoPEMBlock},
-		{"EC private key file", "erin", "ecdsa-sha2-nistp256", []string{"sha256", "sha512"}, ErrFailedPEMParsing},
+		{"existing, but invalid file", "demo.layout.template", "ecdsa-sha2-nistp521", []string{"sha512"}, ErrNoPEMBlock},
+		{"EC private key file", "erin", "ecdsa-sha2-nistp521", []string{"sha256", "sha512"}, ErrFailedPEMParsing},
```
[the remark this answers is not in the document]
shibumi: Erin is a pure EC key and not a ECDSA Key in PKCS8 format.. that's the difference. Such keys are just not supported by the go libraries parsing the PKCS8 keys.

in_toto/util_test.go
```diff
@@ -102,6 +102,7 @@ func TestSubsetCheck(t *testing.T) {
 		{[]string{"sha256"}, NewSet("sha256", "sha512"), true},
 		{[]string{"sha512"}, NewSet("sha256"), false},
 		{[]string{"sha256", "sha512"}, NewSet("sha128", "sha256", "sha512"), true},
+		{[]string{"sha256", "sha512", "sha384"}, NewSet("sha128"), false},
```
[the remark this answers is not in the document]
shibumi: Test coverage, I think. Hard to say.. I should have commented it :S

in_toto/keylib.go
shibumi: Yes, I agree with you that this part was very defensive. I just wanted to make sure that this will run in a graceful error if somebody has ever the brilliant idea to remove sha256 support. :D

in_toto/hashlib.go
shibumi: Do you have a nice name for it? `sha256String` maybe?

in_toto/util_test.go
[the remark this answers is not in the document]
lukpueh: Oh yeah right to cover...
https://github.com/in-toto/in-toto-golang/blob/c4cc4d58b0dfebd29a58abea3d6361ab8a54b808/in_toto/util.go#L138-L140

in_toto/hashlib.go
lukpueh: Hm. I probably would have used `SHA256` but all-caps does not seem idiomatic for go consts. Let's just keep the strings. :)

in_toto/keylib.go
lukpueh: I'm not a big fan of that kind of defensive programming. I think tests are a way better strategy to defend against such brilliant ideas. :)

shibumi:
@lukpueh I have fixed all of the issues mentioned above. Please verify :)

lukpueh (approved):
Thanks for addressing all my comments! LGTM now. One last request, please remove the corresponding todo item from the README.

shibumi:
@lukpueh done :)

shibumi:
@lukpueh can you put a `hacktoberfest` tag on the PR? :D

SantiagoTorres:
@SantiagoTorres I think I just did. I hope it works :)

shibumi:
@SantiagoTorres thanks, but looks like I made a mistake and the correct label name is `hacktoberfest-accepted`. https://hacktoberfest.digitalocean.com/faq

SantiagoTorres:
:)?

shibumi:
@SantiagoTorres should be fine now, I guess:
```
If the maintainer has marked your pull request as invalid, or you have contributed to a repository that isn’t classified with a Hacktoberfest topic, your pull request will not count toward the challenge. Your pull requests will only count toward your participation if they are in a repository with the hacktoberfest topic and once they have been merged, approved by a maintainer or labeled as hacktoberfest-accepted. Additionally, any pull request with the hacktoberfest-accepted label, submitted to any public GitHub repository, with or without the hacktoberfest topic, will be considered valid for Hacktoberfest.
```

PR #91, opened 2021-02-08
Add misc fixes based on go lint

adityasaky:
- Remove underscore prefix from private function
- Make exposed function private
- Id -> ID
- Json -> JSON
- Fix error strings
- Add docstrings for public members

With these fixes, the only issue I'm seeing in `golint` is the package name.


shibumi:
lgtm

PR #93, opened 2021-02-17
In toto master

mikhailswift:

**Fixes issue #**:

requirements**:


in_toto/cmd/run.go
```diff
@@ -0,0 +1,83 @@
+package cmd
```
shibumi, line 1: Just to make sure. The run.go is intended as POC, right? Or do you plan to write/ship an in-toto go client?

mikhailswift: I opened this and #94 in this repo versus our fork accidentally, sorry about that.

Currently I'm doing POC work aimed at using PKI to sign and verify metadata.  Go is much more comfortable to me than python so I chose to fork this project and work through the idea and in the process we threw together a quick CLI to work with it.

PR #101, opened 2021-04-23
Added support for ITE-5 signature generation

kommendorkapten:

ITE-5/SSl siging spec implementation for Go.

**Fixes issue #**:
Adds support for signing and verifying ITE-5 envelopes.

The signing spec is added in a new package `ssl` that is places parallel to the `in_toto` package. What could be discussed is if this package should have a better name, `ssl` can be an overloaded term.
The implementation does not care about the actual signature implementation, clients are expected to inject a signature implementation via dependency injection.

Related to https://github.com/secure-systems-lab/signing-spec/issues/29
I tried to re-create the signature in the [envelope.md](https://github.com/secure-systems-lab/signing-spec/blob/master/protocol.md) but did not get a match, I then ran the jupyter notebook and was not able to get the same signature either, but the signature I got was exatly the same I got in this go PR (see expected signature in `TestEcdsaSign` test case). The issue contains the modified Python code to re-create the signature mentioned in `envelope.md`

requirements**:

- [*] Tests have been added for the bug fix or new feature


trevrosen (approved):

pkg/ssl/sign.go
```diff
@@ -0,0 +1,260 @@
...
+	}
+	if err = binary.Write(&buf, binary.LittleEndian, uint64(l)); err != nil {
+		return nil, err
+	}
+
+	for _, b := range data {
+		var bw int
+		l = len(b)
+
+		if !verifyPaeLength(int64(l)) {
```
dlorenc, line 69: Do we need to do this check? I think the lengths all come from len() which probably guarantees it will be positive:

https://golang.org/pkg/builtin/#len

pkg/ssl/sign.go
```diff
@@ -0,0 +1,260 @@
...
+type Signature struct {
+	KeyID string `json:"keyid"`
+	Sig   string `json:"sig"`
+}
+
+/*
+Pae implementes PASETO Pre-Authentic Encoding
+https://github.com/paragonie/paseto/blob/master/docs/01-Protocol-Versions/Common.md#authentication-padding
+*/
+func Pae(data [][]byte) ([]byte, error) {
```
dlorenc, line 50: Would this ever have more than two elements (Type and Data)? Would it be cleaner to do:

`Pae(payloadType string, payload []byte)`?

pkg/ssl/sign.go
```diff
@@ -0,0 +1,260 @@
...
+		if !verifyPaeLength(int64(l)) {
+			return nil, fmt.Errorf("length must be less than 2^63")
+		}
+		if err = binary.Write(&buf, binary.LittleEndian, uint64(l)); err != nil {
+			return nil, err
+		}
+		if bw, err = buf.Write(b); err != nil {
+			return nil, err
+		}
+		if bw != l {
```
dlorenc, line 78: I think you can skip this check, the writer is guaranteed to write the full amount (or return an error): https://golang.org/pkg/io/#Writer

dlorenc (commented):
Nice job!

pkg/ssl/sign.go
kommendorkapten: Good catch, I'll remove it.

pkg/ssl/sign.go
kommendorkapten: Yeah, you are correct; from the spec: `0 <= len(s) <= cap(s)`. I'll remove the checks.

pkg/ssl/sign.go
kommendorkapten: In the case of SSL or ITE-5, no. But the PAE spec allows for arbitrary number of arguments, so I was honouring that here.
The function is called from from SSL envelope signer (and verifier) with the signature `func (es *EnvelopeSigner) SignPayload(payloadType string, body []byte) (*Envelope, error)` which is very similar to what you want. So from a client to SSL signing spec's view, the interface is `(payloadType string, payload []byte)`. 
And this is even simplified even further from in-toto's view (ITE-5 which mandates a constant payload type), so from a client to in-toto the interface is `func (s *SslSigner) SignPayload(body []byte) (*ssl.Envelope, error)`.

dlorenc: Ah, makes sense. Thanks!

in_toto/model_test.go
```diff
@@ -1509,3 +1513,62 @@ func TestValidatePublicKey(t *testing.T) {
...
+type nilsigner int
+
+func (n nilsigner) Sign(data []byte) ([]byte, string, error) {
+	return data, "nil", nil
+}
+
+func (n nilsigner) Verify(keyID string, data, sig []byte) (bool, error) {
+
+	if keyID == "nil" {
+		var same = len(data) == len(sig)
```
shibumi, line 1526: Am I missing something here or could we just write: `if len(data) != len(sig)`?

in_toto/model.go
```diff
@@ -843,3 +851,37 @@ func (mb *Metablock) Sign(key Key) error {
...
 }
+
+/*
+SslSigner provides signature generation and validation based on the SSL
+Signing Spec: https://github.com/secure-systems-lab/signing-spec
+as describe by: https://github.com/MarkLodato/ITE/tree/media-type/ITE/5
+It wraps the generic SSL envelope signer and enforces the correct payload
+type both during signature generation and validation.
+*/
+type SslSigner struct {
```
shibumi, line 862: nitpick: I think it's fine to use SSLSigner instead of SslSigner. For example the standard library uses OSCPResponse instead of OscpResponse.

shibumi: Or are you afraid this could be mistaken with SSL/TLS?

shibumi: Same for PAE btw :)

in_toto/model.go
```diff
@@ -61,6 +66,9 @@ var ErrNoPublicKey = errors.New("the given key is not a public key")
 // for example: curve size = "521" and scheme = "ecdsa-sha2-nistp224"
 var ErrCurveSizeSchemeMismatch = errors.New("the scheme does not match the curve size")
 
+// ErrInvalidPayloadType indicates that the envelope used an unkown payload type
+var ErrInvalidPayloadType = fmt.Errorf("unknown payload type")
```
shibumi, line 70: Is there a reason for using fmt.Errorf instead of errors.New here? You don't seem to need the formatting + if we use errors.New() we can still wrap additional information around it via: fmt.Errorf("%w: %s", ErrInvalidPayloadType, someUsefulInformation).

See also: https://blog.golang.org/go1.13-errors

pkg/ssl/sign.go
```diff
@@ -0,0 +1,260 @@
...
+func Pae(data [][]byte) ([]byte, error) {
+	var buf = bytes.Buffer{}
+	var l = len(data)
+	var err error
+
+	// Negative value sets the highest bit to 1.
+	// With two complements encoding, high bit must be set to 0 for
+	// interoperability with languages that lack unsigned integer types.
+	if !verifyPaeLength(int64(l)) {
+		return nil, fmt.Errorf("length must be less than 2^63")
```
shibumi, line 59: Do you think it would make sense to make an error out of this?

something like: `var ErrInvalidLength = errors.New("pae length must be less than 2^63")

same in line 70

pkg/ssl/sign.go
```diff
@@ -0,0 +1,260 @@
...
+			return nil, fmt.Errorf("length must be less than 2^63")
+		}
+		if err = binary.Write(&buf, binary.LittleEndian, uint64(l)); err != nil {
+			return nil, err
+		}
+		if bw, err = buf.Write(b); err != nil {
+			return nil, err
+		}
+		if bw != l {
+			return nil, fmt.Errorf("failed to write all bytes")
```
shibumi, line 79: would be cool to make an own error from this as well. This way we can export this error and document it :)

pkg/ssl/sign.go
```diff
@@ -0,0 +1,260 @@
...
+	var providers []SignVerifier
+
+	for _, sv := range p {
+		if sv != nil {
+			providers = append(providers, sv)
+		}
+	}
+
+	if len(providers) == 0 {
+		return nil, fmt.Errorf("no signers provided")
```
shibumi, line 147: Same here as well. An error variable would give us more documentation.

shibumi:
I've added a few nitpicks :)

in_toto/model.go
adityasaky: > Or are you afraid this could be mistaken with SSL/TLS?

This needn't be a concern, as we're working on renaming the singing-spec. https://github.com/secure-systems-lab/signing-spec/issues/16

kommendorkapten:
@shibumi thank you. I'll go over your comments today.

pkg/ssl/sign.go
kommendorkapten: This code is now removed based on this comment: https://github.com/in-toto/in-toto-golang/pull/101#discussion_r624705233
So only the original error is returned when all bytes could not be returned.

pkg/ssl/sign.go
kommendorkapten: This code is removed now as per comment https://github.com/in-toto/in-toto-golang/pull/101#discussion_r624704390
The condition that triggers `verifyPaeLength`to return false can not happen with the current implementation.

pkg/ssl/sign.go
kommendorkapten: Yes, makes sense.

in_toto/model.go
kommendorkapten: No, just an old habit. Errors created via `fmt.Errorf` can be wrapped too, as if no error is provided as an argument `fmt.Errorf` is just a pass-through to `errors.New()`. But as we don't need the formatting here I'll switch 👍

pkg/ssl/sign.go
kommendorkapten: Agree. Fixed!

in_toto/model.go
kommendorkapten: Yeah, SSL would be more in line, and as I think I mentioned somewhere in the PR "SSL" is a slightly overloaded term so a change would be welcomed :)

in_toto/model_test.go
kommendorkapten: No, I think I initially had a more complex implementation in mind, but it was scrapped but I forgot to clean up all of it :). I'll fix.

shibumi (approved):

shibumi:
Looks good to me now! :)

pkg/ssl/sign_test.go
```diff
@@ -0,0 +1,515 @@
...
+	"github.com/codahale/rfc6979"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestPAE(t *testing.T) {
+	t.Run("Nil", func(t *testing.T) {
+		var want = make([]byte, 8)
+
+		got, err := PAE(nil)
+		assert.Nil(t, err, "Unexpectted error")
```
adityasaky, line 21:
```suggestion
		assert.Nil(t, err, "Unexpected error")
```

pkg/ssl/sign_test.go
```diff
@@ -0,0 +1,515 @@
...
+		assert.Equal(t, want, got, "Wrong encoding")
+	})
+	t.Run("Empty", func(t *testing.T) {
+		var want = make([]byte, 8)
+
+		got, err := PAE([][]byte{})
+		assert.Nil(t, err, "Unexpectted error")
```
adityasaky, line 28:
```suggestion
		assert.Nil(t, err, "Unexpected error")
```

pkg/ssl/sign_test.go
```diff
@@ -0,0 +1,515 @@
...
+		assert.Equal(t, want, got, "Wrong encoding")
+	})
+	t.Run("['']", func(t *testing.T) {
+		var want = []byte{0x01, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00}
+
+		got, err := PAE([][]byte{[]byte("")})
+		assert.Nil(t, err, "Unexpectted error")
```
adityasaky, line 35:
```suggestion
		assert.Nil(t, err, "Unexpected error")
```

pkg/ssl/sign_test.go
```diff
@@ -0,0 +1,515 @@
...
+		assert.Equal(t, want, got, "Wrong encoding")
+	})
+	t.Run("['test']", func(t *testing.T) {
+		var want = []byte{0x01, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x04, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x74, 0x65, 0x73, 0x74}
+
+		got, err := PAE([][]byte{[]byte("test")})
+		assert.Nil(t, err, "Unexpectted error")
```
adityasaky, line 42:
```suggestion
		assert.Nil(t, err, "Unexpected error")
```

pkg/ssl/sign_test.go
```diff
@@ -0,0 +1,515 @@
...
+			0x78, 0x61, 0x6d, 0x70, 0x6c, 0x65, 0x2e, 0x63,
+			0x6f, 0x6d, 0x2f, 0x48, 0x65, 0x6c, 0x6c, 0x6f,
+			0x57, 0x6f, 0x72, 0x6c, 0x64, 0x0b, 0x00, 0x00,
+			0x00, 0x00, 0x00, 0x00, 0x00, 0x68, 0x65, 0x6c,
+			0x6c, 0x6f, 0x20, 0x77, 0x6f, 0x72, 0x6c, 0x64,
+		}
+
+		got, err := PAE([][]byte{[]byte("http://example.com/HelloWorld"),
+			[]byte("hello world")})
+		assert.Nil(t, err, "Unexpectted error")
```
adityasaky, line 59:
```suggestion
		assert.Nil(t, err, "Unexpected error")
```

pkg/ssl/sign_test.go
```diff
@@ -0,0 +1,515 @@
...
+			Signature{
+				KeyID: keyID,
+				Sig:   "Cc3RkvYsLhlaFVd+d6FPx4ZClhqW4ZT0rnCYAfv6/ckoGdwT7g/blWNpOBuL/tZhRiVFaglOGTU8GEjm4aEaNA==",
+			},
+		},
+	}
+
+	signer, _ := NewEnvelopeSigner(ecdsa)
+	env, err := signer.SignPayload(payloadType, []byte(payload))
+	assert.Nil(t, err, "unexecpted error")
```
adityasaky, line 310:
```suggestion
	assert.Nil(t, err, "unexpected error")
```

adityasaky (commented):
Some minor typos aside, this looks good to me. I did have the same question as @shibumi re the use of `same`, but that's also a minor nit.

Thank you, @kommendorkapten!

kommendorkapten:
Thank you @adityasaky , I'll fix the spelling errors today. The `same` comment was fixed in a commit a few days ago, but the conversation is not marked as resolved.

shibumi:
@adityasaky's feedback got merged. I think we are ready to merge.

shibumi:
@SantiagoTorres  last  comment? :D

adityasaky:
> The same comment was fixed in a commit a few days ago, but the conversation is not marked as resolved.

I think there were more than a single instance, but again I think it's not a problem... :)

https://github.com/in-toto/in-toto-golang/pull/101/files#diff-e09c3bb0624f8914dd9ed3ddb7dddf578a9e6d4ff8504b446d51c71ee1250920R73

kommendorkapten:
@adityasaky Oh, good catch, I missed them. I did a grep for `same` but obviously in the wrong directory :) I'll go over them again!

shibumi:
I spoke with @SantiagoTorres, I think we can merge this. Thanks a lot for your contribution @kommendorkapten  :) nice work!

PR #103, opened 2021-05-05
Add custom CWD for inspections

SantiagoTorres:
*Please fill in the fields below to submit a pull request.  The more
information that is provided, the better.*

**Fixes issue #**: #46

**Description of pull request**: This is a followup after further requests for the feature. In short, the option to select a local directory from where the inspections are run may help keeping things tidy at the time of executing them. Worthwhile mentioning, a stronger security mechanism (i.e., an actual sandbox) may be desirable in the long-run...

 I'm frankly amazed that this still merges 2 years after the fact, so I'd like to review and make sure there are no breaking changes that aren't immediately obvious.

Kudos to @radu-matei for this. I'm just the messenger.

**Please verify and check that the pull request fulfills the following
requirements**:

- [ ] Tests have been added for the bug fix or new feature
- [x] Docs have been added for the bug fix or new feature

Issue #46, opened 2019-08-16
Allow passing directory for inspections


radu-matei:
Currently, you cannot run the inspections when the materials and products are in a different directory than the current directory. 

https://github.com/in-toto/in-toto-golang/blob/b2f1d5db81b09e5daab2f20976d9538c1604ccdc/in_toto/verifylib.go#L39

This means that while you can pass the paths for the layout and root key, you have to be in the same directory with the rest of the materials to execute the verification with this library, or copy around the materials.

Given the fact that most commonly, the in-toto artifacts are usually grouped in the same directory, would it make sense to be able to pass the path?

[This branch](https://github.com/in-toto/in-toto-golang/compare/master...radu-matei:inspections-dir) introduces that path to the `RunInspections` functions, but does not break compatibility for the `InTotoVerify` function by creating a new one, `InTotoVerifyWithDirectory`.

Of course, the implementation is open for discussion, but is the functionality something desired in the library?

Thanks!

lukpueh:
Thanks for the request, @radu-matei!  We have discussed this feature on the Python reference implementation repo a while ago (see in-toto/in-toto#258). 

Back then we concluded that it is reasonable to ask the verifier to copy and change to a temporary directory. This is what the folks at datadog, who had asked for this feature, ended up doing (see https://github.com/DataDog/integrations-core/pull/3282). Would that work for you as well?

On a related side-note, there is an open feature request for parametrizing the writing of inspection link metadata files (see https://github.com/in-toto/in-toto/issues/260).

(cc'ing @SantiagoTorres and @trishankatdatadog)

radu-matei:
I guess that's reasonable - in any case, [this branch](https://github.com/in-toto/in-toto-golang/compare/master...radu-matei:inspections-dir) has the changes required to change the directory, if anyone else needs it in the future.
 
For my use case I ended up directly using a container to perform verifications - see https://github.com/engineerd/in-toto-container.

We can safely close this, thanks!

SantiagoTorres:
Reopening this, it seems that this may be a desirable feature for a couple of folks...

adityasaky:
Closed by #105.

PR #104, opened 2021-05-12
Add custom CWD for inspections

SantiagoTorres:

**Fixes issue #**: #46

**Description of pull request**: This is a followup after further requests for the feature. In short, the option to select a local directory from where the inspections are run may help keeping things tidy at the time of executing them. Worthwhile mentioning, a stronger security mechanism (i.e., an actual sandbox) may be desirable in the long-run...

 I'm frankly amazed that this still merges 2 years after the fact, so I'd like to review and make sure there are no breaking changes that aren't immediately obvious.

Kudos to @radu-matei for this. I'm just the messenger.

This is quite literally #103 + 1 commit since I can't fix the linter errors on the other branch (this is the first time I'm not allowed to commit to PR's of in-toto-golang :/).

requirements**:


SantiagoTorres:
I like this, I think it'd be nice to add tests for the feature. Perhaps to cover:

1. Path is not writable (EACCESS) /does not exist (ENOENT)
2. Path overwrites content (not sure what the behavior should be in this case)
3. Path is a symlink (I wonder what'd happen if somebody can trigger an inspection w/ different privs and we end up writing stuff somewhere nasty...)

What does everybody think?

radu-matei (approved):
LGTM, thanks for picking this up, @SantiagoTorres!

I see the code coverage decreased, do we want to add some tests before merging?

shibumi:
@radu-matei  checkout the 'real' PR: https://github.com/in-toto/in-toto-golang/pull/105 . Closing this

PR #105, opened 2021-05-15
inspections runDir

shibumi:
This PR is a copy of @SantiagoTorres's PR 104 and will be my workbench for the next commits over the next days/week.
The original text can be found here:
---

**Fixes issue #**: #46

**Description of pull request**: This is a followup after further requests for the feature. In short, the option to select a local directory from where the inspections are run may help keeping things tidy at the time of executing them. Worthwhile mentioning, a stronger security mechanism (i.e., an actual sandbox) may be desirable in the long-run...

 I'm frankly amazed that this still merges 2 years after the fact, so I'd like to review and make sure there are no breaking changes that aren't immediately obvious.

Kudos to @radu-matei for this. I'm just the messenger.

This is quite literally #103 + 1 commit since I can't fix the linter errors on the other branch (this is the first time I'm not allowed to commit to PR's of in-toto-golang :/).

requirements**:


shibumi:
Further todo's (stolen from #104 ):


- [ ] Path is not writable (EACCESS) /does not exist (ENOENT)
- [ ] Path overwrites content (not sure what the behavior should be in this case)
- [ ] Path is a symlink (I wonder what'd happen if somebody can trigger an inspection w/ different privs and we end up writing stuff somewhere nasty...)

shibumi:
@SantiagoTorres  test coverage is fine again. I will have a look on your three todos now :)

shibumi:
@SantiagoTorres  I have implemented two of your three todos. Do we want to speak about the second one? I don't know if it's smart to overwrite content. I could just check if the directory is empty and if not fail. What do you think?

SantiagoTorres:
Nice stuff! I wonder if we want to default to "warn if the inspections dir is not empty" and finalize this PR :)

shibumi:
@SantiagoTorres  sounds good. I will check the dir for emptyness + add a few additional tests for the coverage and finalize this PR.

shibumi:
Waiting for Github Actions.. and then this is ready to merge, I think.

shibumi:
@SantiagoTorres  @adityasaky  Do you want to have a final look? We have a slight code coverage decrease, but I think we should get this merged. I have added tests for most edge cases, the code coverage decrease comes from the additional inTotoVerify method. Maybe we should track this in a different issue and schedule a next release. What do you think?

in_toto/util_darwin.go
```diff
@@ -0,0 +1,11 @@
+package in_toto
```
adityasaky, line 1: Curious, but could we combine this and `_linux` into one file, and use a `!windows` constraint instead of maintaining two copies of the same function?

adityasaky: Apart from that, I think this is all fine. :smile:

shibumi: The `!windows` constraint is file based.. we would have two different files again. One for Linux and one for Windows. 

Or am I just not understanding what you mean?

adityasaky: Yes, but currently we have three, `_linux`, `_darwin`, and `_windows`. The first two are identical as well. We can use a single file there with the `!windows` constraint?

shibumi: ah, this is what you mean. Yes, this is possible! Maybe even _unix works as suffix.

shibumi: @adityasaky  check the latest commit.. fixed

adityasaky:
This looks fine to me, thanks @shibumi!

PR #109, opened 2021-05-26
added small example to the readme

omerlh:
I just spent some time digging in the tests to figure how to use this lib, so I added a small example to the readme - to help others.

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,7 +20,7 @@ and navigate to `localhost:8080/pkg/github.com/in-toto/in-toto-golang/in_toto/`
 A very simple example, just to help you starting:
 
 ```go
-package main
+**package main
 
 import (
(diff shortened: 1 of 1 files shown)


shibumi:
This is really cool @omerlh. Very good idea.

Maybe we should include such examples in our code documentation as well. Go allows examples in the "docstrings", IIRC.

omerlh:
Yes please! I have full example with steps for a typical nodeJS app that I can share

shibumi:
@omerlh  Maybe we should have a dedicated docs or examples directory. @SantiagoTorres  what do you think?

SantiagoTorres:
I like this! yeah I think we should have an examples dir...

I'll merge this as-is as part of the README. Thank you @omerlh @shibumi!

PR #114, opened 2021-06-19
Simplify the EnvelopeSigner interface a bit.

dlorenc:
It was returning (ok, err), which means clients have to remember to check
two things. This feels a little less foot-gun prone, and better aligns
with the crypto stdlib.

**Fixes issue #**:

requirements**:

### DIFF
--- a/pkg/ssl/sign_test.go
+++ b/pkg/ssl/sign_test.go
@@ -5,6 +5,7 @@ import (
 	"crypto/elliptic"
 	"crypto/sha256"
 	"encoding/base64"
+	"errors"
 	"fmt"
 	"math/big"
 	"testing"
@@ -12,6 +13,8 @@ import (
 	"github.com/codahale/rfc6979"
 	"github.com/stretchr/testify/assert"
 )
+
+var errLength = errors.New("invalid length")
 
 func TestPAE(t *testing.T) {
 	t.Run("Nil", func(t *testing.T) {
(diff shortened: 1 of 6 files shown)


dlorenc:
cc @kommendorkapten wdyt about this one?

shibumi:
Hi @dlorenc  thanks for your PR.

@SantiagoTorres do we want to get this into v0.2.0 or should this be merged after our 0.3.0 release? IMHO it makes sense to merge this before the next release.

pkg/ssl/sign_test.go
```diff
@@ -67,22 +68,22 @@ func (n nilsigner) Sign(data []byte) ([]byte, string, error) {
 	return data, "nil", nil
 }
 
-func (n nilsigner) Verify(keyID string, data, sig []byte) (bool, error) {
+func (n nilsigner) Verify(keyID string, data, sig []byte) error {
 
 	if keyID == "nil" {
 		if len(data) != len(sig) {
-			return false, nil
+			return errors.New("invalid length")
```
shibumi, line 75: It might make sense to make error types out of these static error strings if we use them more than once.
For example:

`var ErrInvalidPayloadLength = errors.New("invalid payload length")`

For example like this: https://github.com/in-toto/in-toto-golang/blob/master/pkg/ssl/sign.go#L15L23

This way we can add more more documentation to errors (if needed). More Information can be found here: https://blog.golang.org/go1.13-errors

pkg/ssl/sign_test.go
```diff
@@ -67,22 +68,22 @@ func (n nilsigner) Sign(data []byte) ([]byte, string, error) {
...
 		}
 
 		for i := range data {
 			if data[i] != sig[i] {
-				return false, nil
+				return errors.New("invalid signature")
```
shibumi, line 80: Same here

kommendorkapten:
@dlorenc Hey. Nice work, agreed it will simplify things 👍

adityasaky (approved):
LGTM, thanks @dlorenc!

pkg/ssl/sign_test.go
dlorenc: All set! These are tests so I left the errs private rather than exposing them outside the package.

pkg/ssl/sign_test.go
dlorenc: Thanks!

adityasaky:
Thanks again, @dlorenc. @shibumi I think we can sneak this change into 0.2.0 and tag soon?

shibumi:
@adityasaky  it's a minor change, I don't think we need to mention this in our changelog. But it should be in our 0.2.0 release.

adityasaky:
I think our changelog handles this already, from the perspective of 0.2.0. :)

PR #117, opened 2021-07-29
make .metadata optional in Provenance statement

developer-guy:
Signed-off-by: Batuhan Apaydın <batuhan.apaydin@trendyol.com>

**Fixes issue #**:

We (I and @Dentrax) saw the `.metadata` section is defined as an optional field in the [fields](https://github.com/in-toto/attestation/blob/main/spec/predicates/provenance.md#fields) section of the Provenance documentation, so we should add the `omitempty` tag to it too. This PR aims to fix this problem.

requirements**:

### DIFF
--- a/in_toto/model.go
+++ b/in_toto/model.go
@@ -940,7 +940,7 @@ type ProvenanceMaterial struct {
 type ProvenancePredicate struct {
 	Builder   ProvenanceBuilder    `json:"builder"`
 	Recipe    ProvenanceRecipe     `json:"recipe"`
-	Metadata  ProvenanceMetadata   `json:"metadata"`
+	Metadata  *ProvenanceMetadata  `json:"metadata,omitempty"`
 	Materials []ProvenanceMaterial `json:"materials,omitempty"`
 }
 
--- a/in_toto/model_test.go
+++ b/in_toto/model_test.go
@@ -1591,7 +1591,7 @@ func TestDecodeProvenanceStatement(t *testing.T) {
 				DefinedInMaterial: new(int),
 				EntryPoint:        "build.yaml:maketgz",
 			},
-			Metadata: ProvenanceMetadata{
+			Metadata: &ProvenanceMetadata{
 				BuildStartedOn: &testTime,
 				Completeness: ProvenanceComplete{
 					Environment: true,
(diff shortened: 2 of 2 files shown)


adityasaky (changes requested):
Hi @developer-guy, thanks for the patch! I'm not 100% certain this fixes it the way we want it to, though, because it's a struct field. I think we want to also change `Metadata ProvenanceMetadata` to `Metadata *ProvenanceMetadata`, because the struct itself doesn't have an empty value. If we want the default values within `ProvenanceMetadata` I suppose this is okay, though I haven't followed this into the other structs. :)

See: https://play.golang.org/p/ijyRb-By9sN here, the metadata field is still printed, while in https://play.golang.org/p/x0ajUdtSabJ it is not.

developer-guy:
> Hi @developer-guy, thanks for the patch! I'm not 100% certain this fixes it the way we want it to, though, because it's a struct field. I think we want to also change `Metadata ProvenanceMetadata` to `Metadata *ProvenanceMetadata`, because the struct itself doesn't have an empty value. If we want the default values within `ProvenanceMetadata` I suppose this is okay, though I haven't followed this into the other structs. :)
> 
> See: https://play.golang.org/p/ijyRb-By9sN here, the metadata field is still printed, while in https://play.golang.org/p/x0ajUdtSabJ it is not.

thank you so much for warning us @adityasaky, we forgot to do it, but now it should be ok, I updated the code according to your changes, thank you so much 🥳🙋🏻‍♂️

adityasaky:
Looks like the tests need to be updated to match the change in field. :)

developer-guy:
> Looks like the tests need to be updated to match the change in field. :)

fixed, thank you so much again, sorry about all the mistakes 🤦🏻‍♂️

adityasaky:
Thanks @developer-guy, and no problem at all. Before we merge, I just want to check with / give a heads up to @dlorenc and @SantiagoTorres. We don't want to break something without warning...:)

developer-guy:
okay, make sense @adityasaky 🤝🙋🏻‍♂️

dlorenc:
+1 from me!

SantiagoTorres:
It also LGTM :)

adityasaky:
Thanks!

adityasaky:
And thanks again, @developer-guy! :)

developer-guy:
Thank you so much folks!! ❤️🙋🏻‍♂️

PR #119, opened 2021-08-24
ITE7 Prototype Implementation

mikhailswift:
This is a prototype implementation of ITE7 as drafted and proposed at:
https://github.com/in-toto/ITE/pull/21

* Adds the concept of certificate constraints.  This allows layout
  creators to define the shape of a X509 certificate that is allowed to
  act as a functionary for each step of a layout.
* Adds roots and intermediate CAs to the layout.  This allows layout
  creators to define the chain of trust that a X509 certificate must
  fall within to act as a functionary.
* Adds the leaf certificate to signatures.  This will need to be updated
  to be compatible with DSSE if ITE5 is accepted into the specficiation.

**Fixes issue #**:

requirements**:


mikhailswift:
We're still working on splitting up our fork into separate pull requests.  We're thinking two more will be opened: one for the CLI and another for the SPIFFE/SPIRE plugin.

We created end-to-end testing for all of this functionality but it relies upon the CLI pull request.  In the meantime I'll continue to work and add unit tests and refining the documentation.

in_toto/runlib.go
```diff
@@ -299,14 +309,71 @@ func InTotoRun(name string, runDir string, materialPaths []string, productPaths
...
 	if !reflect.ValueOf(key).IsZero() {
 		if err := linkMb.Sign(key); err != nil {
 			return linkMb, err
 		}
 	}
+
+	return linkMb, nil
+}
+
+func InTotoRecordStart(name string, materialPaths []string, key Key, hashAlgorithms, gitignorePatterns []string, lStripPaths []string) (Metablock, error) {
```
shibumi, line 327: Please add a brief description in form of docstrings for InTotoRecordStart/Stop.

in_toto/runlib.go
```diff
@@ -299,14 +309,71 @@ func InTotoRun(name string, runDir string, materialPaths []string, productPaths
...
+		Type:        "link",
+		Name:        name,
+		Materials:   materials,
+		Products:    map[string]interface{}{},
+		ByProducts:  map[string]interface{}{},
+		Command:     []string{},
+		Environment: map[string]interface{}{},
+	}
+
+	if !reflect.ValueOf(key).IsZero() {
```
shibumi, line 344: Why not just a key == nil check on that position? I see this reflect isZero construct the first time. It checks for zero values in the key object right? Like: The key has been initialized but it's empty? Is this correct?

in_toto/model.go
```diff
@@ -750,11 +822,11 @@ func (mb *Metablock) Load(path string) (err error) {
 		mb.Signed = layout
 
 	} else {
-		return fmt.Errorf("The '_type' field of the 'signed' part of in-toto" +
+		return fmt.Errorf("the '_type' field of the 'signed' part of in-toto" +
 			" metadata must be one of 'link' or 'layout'")
 	}
 
-	return nil
+	return jsonFile.Close()
```
shibumi, line 829: Shouldn't that jsonFile.Close() be called anyway due to the defer jsonFile.Close() at beginning? Wouldn't we call it twice?

in_toto/keylib_test.go
```diff
@@ -16,17 +16,17 @@ func TestLoadKey(t *testing.T) {
...
-		{"ecdsa private key (P521)", "frank", "ecdsa-sha2-nistp521", []string{"sha256", "sha512"}, "434cf7c5b168f6ea4c7e6e67afa74a02625310530f1664f761637bdc7ad8f8df"},
-		{"ecdsa public key (P521)", "frank.pub", "ecdsa-sha2-nistp521", []string{"sha256", "sha512"}, "434cf7c5b168f6ea4c7e6e67afa74a02625310530f1664f761637bdc7ad8f8df"},
-		{"ecdsa private key (P384)", "grace", "ecdsa-sha2-nistp384", []string{"sha256", "sha512"}, "a5522ebccd492f64e6ec0bbcb5eb782708f6e26709a3712e64fff108b98e5142"},
-		{"ecdsa public key (P384)", "grace.pub", "ecdsa-sha2-nistp384", []string{"sha256", "sha512"}, "a5522ebccd492f64e6ec0bbcb5eb782708f6e26709a3712e64fff108b98e5142"},
-		{"ecdsa private key (P224)", "heidi", "ecdsa-sha2-nistp224", []string{"sha256", "sha512"}, "fae849ef9247cc7d19ebd33ab63b5d18a31357508fd82d8ad2aad6fdcc584bd7"},
-		{"ecdsa public key (P224)", "heidi.pub", "ecdsa-sha2-nistp224", []string{"sha256", "sha512"}, "fae849ef9247cc7d19ebd33ab63b5d18a31357508fd82d8ad2aad6fdcc584bd7"},
+		{"rsa public key", "alice.pub", "rsassa-pss-sha256", []string{"sha256", "sha512"}, "d7b728368798278c6bbd43e57b9ff9794be73c24edc574fdaae67efcbc34e23a"},
```
shibumi, line 19: Was generating new pub keys necessary for the PR?

shibumi:
@mikhailswift Thanks for the PR. I have added a few questions. These are not meant as request for changes, they are just open questions on my side :)

in_toto/model.go
shibumi: I see this "defer early & return close"-construct the first time :D if this works, I learned something new. Cool idea.

in_toto/keylib_test.go
mikhailswift: There's a few things going on here, first is that due to adding the `certificate` field to the KeyVal struct all key ids changed.  I had previously attempted to overload the public field and use that for the certificate but it proved difficult to manage.

Alice's key did have to be regenerated due to her private not being checked into the repo.  I needed to re-sign the test data so I needed Alice's private key.

We can revisit the KeyVal struct changes and attempt to find a way to facilitate these changes without changing key ids if wanted.  It's not ideal to have to update all key ids obviously.

in_toto/runlib.go
mikhailswift: Correct, this is checking to see if the value of `key` is the zero value of the `Key` struct.  A nil check won't work here due to `key` being an instance of `Key` and not `*Key`.

An equivalent way to write this would be `!key == Key{}`, which may be a bit more clear.  This was done this way to remain consistent with the check elsewhere in the code base (specifically here https://github.com/in-toto/in-toto-golang/pull/119/files/31b889b97d898f343cb7c6b3c1ef092dbe5f5506#diff-20ae90a40c9494c3e2e4a995a57bc508621a0e473dd74ae811f560d187ce7887R318 )

in_toto/model.go
mikhailswift: It would be called twice in the happy path, yes.  One of my colleagues wrote this bit here and it somewhat raised my eyebrows as well.  I believe the thinking is that if an error occurred in one of the calls in this function such as `checkRequiredJSONFields` and an error also occurred in the deferred `Close`, the deferred close would overwrite the error from `checkRequiredJSONFields`.

in_toto/runlib.go
shibumi: Thank you for the explanation :)  I learned something new.

in_toto/keylib_test.go
shibumi: > Alice's key did have to be regenerated due to her private not being checked into the repo. I needed to re-sign the test data so I needed Alice's private key.

Oh, this must have been my fault sorry. Looks like I forgot adding it to the repository when I was working on the in-toto run implementation.

I know what you mean, but I think getting this merged is priority number one right now + I would prefer better code coverage over the key ID issue.

mikhailswift: I will definitely focus more unit tests today to get that coverage up

adityasaky:
Hi @mikhailswift, thank you for your work on this, this looks great! I was wondering if we could make the new fields optional so that the implementation continues to support non ITE-7 metadata. I'm wary of breaking, say, verification workflows where the metadata is generated using the python implementation but the verification uses the Go implementation, or even where folks have older metadata generated using the Go implementation.

mikhailswift:
Absolutely, 100% in favor of that and will focus on getting that done ASAP :)

adityasaky:
Thank you, I appreciate it!

mikhailswift:
Sorry for the delay in this, the past week has been chaotic.  My latest commit makes the new fields optional.  Coincidentally this also fixes the KeyID problem discussed before.

shibumi (approved):

shibumi:
looks good for me! :)

in_toto/model.go
```diff
@@ -716,24 +716,41 @@ type Metablock struct {
...
+
+		allFields = append(allFields, field)
 	}
 
 	// Assert that there's a key in the passed map for each tag
 	for _, field := range allFields {
-		if _, ok := obj[field]; !ok {
-			return fmt.Errorf("required field %s missing", field)
+		if _, ok := obj[field.name]; !ok && !field.omitempty {
+			return fmt.Errorf("required field %s missing", field.name)
```
mikhailswift, line 753: probably not super important, but this for loop and the one above could be combined so we only have to do one pass over all the fields.

adityasaky:
Thanks @mikhailswift! Looking forward to the other two PRs :)

PR #120, opened 2021-08-31
Add support for Go 1.17 and deprecate Go 1.15

shibumi:
Go 1.17 has been released on the 16th of August 2021. Let's deprecate Go 1.15 and follow the Go release cycle.

### DIFF
--- a/in_toto/util_unix.go
+++ b/in_toto/util_unix.go
@@ -1,3 +1,4 @@
+//go:build linux || darwin || !windows
 // +build linux darwin !windows
 
 package in_toto
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -4,7 +4,7 @@ jobs:
   test:
     strategy:
       matrix:
-        go-version: [1.15.x, 1.16.x]
+        go-version: [1.16.x, 1.17.x]
         os: [ubuntu-latest, macos-latest, windows-latest]
     runs-on: ${{ matrix.os }}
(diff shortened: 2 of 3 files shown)


in_toto/util_unix.go
```diff
@@ -1,3 +1,4 @@
+//go:build linux || darwin || !windows
 // +build linux darwin !windows
```
adityasaky, line 2: According to https://pkg.go.dev/go/build/constraint@go1.16.7, the parser for 1.16 also supports the new go:build constraints. Can we check this out?

adityasaky:
Hey @shibumi, thanks for bumping this. One small check, and we should be good to go!

in_toto/util_unix.go
shibumi: The tests succeed, so I guess this works. The CI just fails because of network issues with coverall..

adityasaky: Oh yes, I'm just wondering if we can get rid of ` // +build linux darwin !windows ` as even 1.16 seems to support the `go:build` format.

shibumi: Ah okay, now I get what you mean... I think this would make our code incompatible with Go < 1.16. I would argue we wait a little bit more.. this string/comment has no effect on future Go versions and we are not losing anything by adding it.

adityasaky: Fair enough.

PR #122, opened 2021-09-10
dsse: transfer the dsse part into a new repo

shibumi:

**Fixes issue #**:

requirements**:


adityasaky:
Thanks for picking this up, @shibumi! Let me cut v0.1.0 of go-sslib though.

adityasaky:
cc @dlorenc for sigstore's dependence on this code.

shibumi:
> Thanks for picking this up, @shibumi! Let me cut v0.1.0 of go-sslib though.
> 
> Here we go: https://github.com/secure-systems-lab/go-securesystemslib/releases/tag/v0.1.0

cutting a new release for it is a good idea. Makes it easier to pinpoint it. Good idea!

We should also start adding Maintainers.md files etc in the go-securesystemslib repository.

in_toto/model.go
```diff
@@ -1070,11 +1070,11 @@ It wraps the generic SSL envelope signer and enforces the correct payload
 type both during signature generation and validation.
 */
 type SSLSigner struct {
```
adityasaky, line 1072: We should rename this and modify its docstring.

shibumi: Renaming the SSLSigner struct would break compatibility to older versions. Do we really want to do this?

adityasaky: https://github.com/golang/go/wiki/Deprecated

shibumi: I think using deprecated is a good idea here.. IMHO the whole struct should be replaced via an interface anyway (for easier test mockups).

adityasaky: A bunch of the patches have been merged now! I think we can proceed with this PR, with `DSSESigner` added in.

Issue #124, opened 2021-09-10
Potential bug in InTotoRun


shibumi:
**Description of issue or feature request**:

https://github.com/in-toto/in-toto-golang/blob/02b98c8d4e2220c5fa5b992f39cd167986cede78/in_toto/runlib.go#L293 

We are not handling an empty `cmdArgs` either as `nil` or empty string slice. We should check if our code still works with both and that we are not running into a panic.

Furthermore, we should add tests to cover these edge cases.

adityasaky:
Closed by #126.

PR #126, opened 2021-09-11
fix: add nil/len guards for RunCommand

shibumi:

**Fixes issue #**: #124

requirements**:

### DIFF
--- a/in_toto/runlib_test.go
+++ b/in_toto/runlib_test.go
@@ -377,6 +377,23 @@ func TestRunCommand(t *testing.T) {
 	}
 }
 
+func TestRunCommandErrors(t *testing.T) {
+	tables := []struct {
+		CmdArgs       []string
+		RunDir        string
+		ExpectedError error
+	}{
+		{nil, "", ErrEmptyCommandArgs},
+		{[]string{}, "", ErrEmptyCommandArgs},
+	}
+	for _, table := range tables {
+		_, err := RunCommand(table.CmdArgs, table.RunDir)
+		if !errors.Is(err, ErrEmptyCommandArgs) {
+			t.Errorf("RunCommand did not provoke expected error. Got: %s, want: %s", err, ErrEmptyCommandArgs)
+		}
+	}
+}
+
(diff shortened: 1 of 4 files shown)


shibumi:
This PR is missing more tests for the InTotoRun function. Right now I am only testing the RunCommand func. We might want to test link creation for an empty CmdArgs parameter as well

in_toto/runlib_test.go
```diff
@@ -360,6 +360,23 @@ func TestRunCommand(t *testing.T) {
 	}
 }
 
+func TestRunCommandErrors(t *testing.T) {
```
adityasaky, line 363: Can / should we also test by invoking `InTotoRun`?

adityasaky: Oh whoops, I missed your comment on the PR.

shibumi: yes we can. I will add a few more tests for this.

shibumi:
> We should add -x / --no-command to the CLI to indicate no commands.

Is it even possible to create an in-toto link attestation without a command being run?

adityasaky:
Yes, the field should just be empty. The reference implementation supports it. Also note that links generates via record always have the command field empty.

shibumi:
Just to be sure:
> We should add -x / --no-command to the CLI to indicate no commands.

The `-x` flag is for the subcommand `run`, right?

in_toto/runlib_test.go
adityasaky: Reviving this one. I see what you mean but I think it's okay to test that the no command flag works as intended and still results in a link, no?

adityasaky: @shibumi added a test checking if link metadata is successfully generated. Also caught that a minor difference in `byProducts` had crept in.

adityasaky:
@shibumi @pxp928 I think this is good to go, please take a look. If we merge this in, we can soon cut a release I think.

pxp928 (approved):
Looks good based on the code.

PR #135, opened 2021-09-29
fix: do not alter content when hashing files

shibumi:
**Fixes issue:**

None

**Description:**
Before we have normalized lines for supporting different
operating systems. The line normalization led to altered
byte segments in BLOBs. This patch disables the line normalization
feature for now, adds a test and removes windows CI runs.

requirements:**


shibumi:
@adityasaky  I forgot one thing.. I switched to 'alice.pub' as test file.. but alice.pub is not a binary.. so the test does not make sense any more. Let me use the test tar.gz file instead  for the test.

shibumi:
Ok, now it's ready for review...

shibumi:
@adityasaky  and me just realized that we have no issues with the foo.tar.gz file, because this file does not have the corresponding byte segments... We have to add another test file that triggers the bug (when line normalization is enabled).

shibumi:
I have added a helloworld binary compiled with Go 1.17 and I can confirm this time the bug gets triggered, because there are windows-like line breaks as byte segments in this file.

shibumi:
@adityasaky  This is ready now. Note to the student who should work on this: If we add another flag to the function we should probably try to toggle it in the test function and test the reproduction of the bug.

adityasaky (approved):
LGTM, thanks @shibumi!

Issue #139, opened 2021-10-01
Left stripping should check if resulting paths are not unique


adityasaky:
**General information:**

* Version (tag or commit): v0.3.2
* Operating system (Linux/Mac/Windows): N/A
* Go version: N/A
* Go build flags (if any): N/A

**Description of the bug:**

When applying left strip patterns to artifacts being recorded, the implementation does not check if the result of left stripping is unique. For example, if `/dir1/foo` and `/dir2/foo` are being recorded, and `/dir1/` and `/dir2/` are part of the left strip patterns, the artifact recorded second will overwrite the hashes recorded for the first. Instead, the implementation should error out to warn users.

The python implementation handles this here: https://github.com/in-toto/in-toto/blob/develop/in_toto/runlib.py#L107-L110.

![image](https://user-images.githubusercontent.com/8928778/135669898-442eee89-305d-48d9-92b7-fda91e72a2ff.png)

**Anything special you want to tell us?**

Nope.

PR #140, opened 2021-10-02
added uniqueness test to resulting paths

kadern0:
**Fixes issue:**
#139 

**Description:**
Adds a check to ensure the resulting artifact name is unique

requirements:**

### DIFF
--- a/in_toto/runlib.go
+++ b/in_toto/runlib.go
@@ -182,7 +182,11 @@ func recordArtifacts(paths []string, hashAlgorithms []string, gitignorePatterns 
 						break
 					}
 				}
-
+				// Check if path is unique
+				_, existingPath := artifacts[path]
+				if existingPath {
+					return fmt.Errorf("Prefix selection has resulted in non unique dictionary key: %s", path)
+				}
 				artifacts[path] = artifact
 				return nil
 			})
--- a/in_toto/runlib_test.go
+++ b/in_toto/runlib_test.go
@@ -295,6 +295,16 @@ func TestRecordArtifacts(t *testing.T) {
 		t.Errorf("RecordArtifacts returned '(%s, %s)', expected '(%s, nil)'",
 			result, err, expected)
 	}
+	// Test duplicated artifact after left strip
+	if err := ioutil.WriteFile("tmpdir/foo.tar.gz", []byte("abc"), 0400); err != nil {
+		t.Errorf("Could not write tmpfile: %s", err)
+	}
+	_, err = RecordArtifacts([]string{"foo.tar.gz",
+		"tmpdir/foo.tar.gz"}, []string{"sha256"}, nil, []string{"tmpdir/"})
+	if err == nil {
+		t.Error("Duplicated path error expected")
+	}
+
 	if err := os.RemoveAll("tmpdir"); err != nil {
 		t.Errorf("Could not remove tmpdir: %s", err)
 	}


shibumi:
LGTM.

I just realized that all of our error messages are upper-case. This is a violation of a go standard about error messages. @adityasaky we might should track this in a separate issue.

We can merge this one here

adityasaky:
We also likely want to make it consistent with the use of `errors` across the board? We can track it in a separate issue, but IMO we can also start by using it in new code or code that's already modifying some error string?

in_toto/runlib.go
```diff
@@ -182,7 +182,11 @@ func recordArtifacts(paths []string, hashAlgorithms []string, gitignorePatterns
 						break
 					}
 				}
-
+				// Check if path is unique
+				_, existingPath := artifacts[path]
+				if existingPath {
+					return fmt.Errorf("Prefix selection has resulted in non unique dictionary key: %s", path)
```
adityasaky, line 188:
```suggestion
					return fmt.Errorf("left stripping has resulted in non unique dictionary key: %s", path)
```

adityasaky (approved):
Minor nit aside, this LGTM @kadern0, thanks! I'd like to hear from @shibumi about switching to `errors` here, but happy to merge if we want to hold off on that.

in_toto/runlib.go
shibumi: I don't know. I think I would prefer to handle the error strings isolated in a new PR. We shouldn't mix code reformatting with bugfixes or features. Commits need to be atomic.

kadern0: I can update the error string and squash both commits if you want.

shibumi: @kadern0 no, not necessary. I really think we should address the wrong error string formatting in a different issue/PR.
Do you agree @adityasaky ?

adityasaky: Error string formatting definitely belongs in its own issue.

adityasaky: To be clear, @kadern0 I think the suggestion I made (or any other version of including "left strip" in the error message for clarity) is sufficient to merge this PR. Apologies for the non PR specific discussions that may have caused some confusion. I'd only like to update the error message to ensure that anyone using the functionality know _exactly_ what the issue is.

shibumi: Ah okay. Thanks for the clarification. Can we use "Left stripping ..." in this PR then + add a new issue for solving the issues with error formatting in general?

adityasaky: I don't think we need to worry about capitalizing "left" only to handle it again in an error strings formatting PR. Since we've (informally) decided to make that change, we can choose to apply it to any changes we make to error strings / new error strings even if the formatting hasn't been applied to the rest. I may have misunderstood you though. :smile:

I also do wonder if we can just handle everything (formatting, using `errors` consistently) under the umbrella of #5.

shibumi: Yep, makes sense to handle this under the umbrella of #5.

> I don't think we need to worry about capitalizing "left" only to handle it again in an error strings formatting PR.

I am a fan of atomic PRs. A PR should consist only of the change necessary for solving the specified problem. Adding reformatting to it adds more "complexitiy". If we merge this PR with one error string starting with lower case, this would be inconsistent with the rest of the code base. It's better to handle this separately, imho. But feel free to merge it.

adityasaky: I'm okay either way :smile:

shibumi: In that case, just let us merge this and do the error string reformat under the umbrella of #5 like you suggested.

shibumi:
Thanks @kadern0 for your work.

Issue #136, opened 2021-09-29
RecordArtifact needs a flag for line normalization


shibumi:
The PR #135 has disabled line normalization due to issues with binary files.
We should introduce a boolean parameter to the function to toggle it.
On default it should be disabled and only enabled when explicitly set to true.

This issue involves altering the RecordArtifact function in runlib.go and adding a new flag for our in-toto CLI.

developer-guy:
look folks what I found 👀😋 @dentrax @erkanzileli

we would love to help @shibumi

shibumi:
@developer-guy oh this issue is supposed to one of @adityasaky @SantiagoTorres's students :) sorry.
But have a look on the mentioned PR and leave your comments on this.

alanssitis:
@shibumi I'll gladly take on the challenge 😄!

shibumi:
@alanssitis cool. Ping me in slack if you have any questions :) happy to help

PR #142, opened 2021-10-04
Add lineNormalization parameter to RecordArtifact

alanssitis:
**Fixes issue:** #136 

**Description:** Introduced boolean parameter lineNormalization to RecordArtifact in runlib.go and added new flag to the CLI. This parameter controls whether contents are normalized. Normalization issue encountered in #135 was not fixed, however. 
*Not yet finished, would need some advice on making CLI flag shorter.*

requirements:**


cmd/in-toto/record.go
```diff
@@ -138,6 +138,14 @@ command is executed. Symlinks are followed.`,
 are stored in the resulting link metadata after the
 command is executed. Symlinks are followed.`,
 	)
+
+	recordCmd.Flags().BoolVarP(
+		&lineNormalization,
+		"line-normalization",
```
adityasaky, line 144:
```suggestion
		"normalize-line-endings",
```

IMO this reads better as a flag. I'm also not sure if we want to include a short flag for this one.

cmd/in-toto/record.go
```diff
@@ -138,6 +138,14 @@ command is executed. Symlinks are followed.`,
...
+		"o",
+		false,
+		`Enable line normalization.`,
```
adityasaky, line 147: Ideally, the help message should talk a bit more about what enabling this flag does, perhaps with a quick example, and when to use it.

cmd/in-toto/run.go
```diff
@@ -121,6 +121,14 @@ of another.`,
 in environment variables or config files. See Config docs for details.`,
 	)
 
+	runCmd.Flags().BoolVarP(
+		&lineNormalization,
+		"line-normalization",
```
adityasaky, line 126: Same here as in record.

adityasaky (changes requested):
This looks like a great start. Can we add some tests where the flag is `true`? We should also start testing on windows again, and see how to enable it for tests there...

cc @shibumi

alanssitis:
Yeah definitely @adityasaky!

adityasaky:
It does occur to me that perhaps we should replace the CLI stuff in the README with a link to the doc folder... cc @shibumi

PR #149, opened 2021-10-08
Configure cobra using standard layout

lukehinds:
This sets cobra up using the standard approach (the default when running `cobra init`)

It's then easier to run the app with commands such as:

`go run main.go run`

Not hugely adamant this gets merged, but figured its worth putting up as quicker than an issue.

Signed-off-by: Luke Hinds <luke@redhat.com>


lukehinds:
> This should be good to go after updating the workflow to point to the new verify script for the docs.
> 
> Thanks @lukehinds!

This should fix it d9ba4ac1a275081b71c8dcf34e3d5e22f08f3e80

adityasaky:
Updated the makefile and moved the verification script to `scripts/` in this PR, @lukehinds. I figured the relatively minor scripts change could sneak in under the umbrella of "use standard templates".

shibumi:
@lukehinds  is there any way to get rid of the date in the documentation? I mean lines like this:
```
###### Auto generated by spf13/cobra on 8-Oct-2021
```

The auto generation notice itself is okay, I just see no reason for the date, if we have a commit date already.

adityasaky:
Added the flag to disable the autogen message and regenerated the docs. :)

PR #153, opened 2021-10-27
Move slsa-provenance predicate format into v0.1 package

priyawadhwa:
**Fixes issue:**
I'm working on merging in a proposal for v0.2 for the `slsa-provenance` predicate format (https://github.com/slsa-framework/slsa/pull/179). Once that's in, I was planning on adding the new version to this repo. There doesn't seem to be a way to version the [ProvenancePredicate](https://github.com/in-toto/in-toto-golang/blob/master/in_toto/model.go#L1033) so I moved the slsa-provenance predicate to a v0.1 directory. 

**Description:**
I moved the predicate model to `pkg/in_toto/slsa_provenance/v0.1`. When adding in new versions, we can create `v0.x` directories as needed. I also named the import to this package in `models.go` as `latest`, so only the import needs to be updated as new versions are added. 

Feedback would be appreciated if someone has a better way to do this! 

requirements:**

### DIFF
--- a/in_toto/model.go
+++ b/in_toto/model.go
@@ -15,7 +15,7 @@ import (
 	"strings"
 	"time"
 
-	latest "github.com/in-toto/in-toto-golang/in_toto/slsa_provenance/v0.1"
+	slsa "github.com/in-toto/in-toto-golang/in_toto/slsa_provenance/v0.1"
 
 	"github.com/secure-systems-lab/go-securesystemslib/dsse"
 )
@@ -961,8 +961,8 @@ func (mb *Metablock) Sign(key Key) error {
 
 // Subject describes the set of software artifacts the statement applies to.
 type Subject struct {
-	Name   string           `json:"name"`
-	Digest latest.DigestSet `json:"digest"`
+	Name   string         `json:"name"`
+	Digest slsa.DigestSet `json:"digest"`
 }
 
 // StatementHeader defines the common fields for all statements
@@ -985,7 +985,7 @@ type Statement struct {
 // ProvenanceStatement is the definition for an entire provenance statement.
 type ProvenanceStatement struct {
 	StatementHeader
-	Predicate latest.ProvenancePredicate `json:"predicate"`
+	Predicate slsa.ProvenancePredicate `json:"predicate"`
 }
 
 // LinkStatement is the definition for an entire link statement.
--- a/in_toto/model.go
+++ b/in_toto/model.go
@@ -15,6 +15,8 @@ import (
 	"strings"
 	"time"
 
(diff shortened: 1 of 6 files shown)


in_toto/model.go
```diff
@@ -15,6 +15,8 @@ import (
 	"strings"
 	"time"
 
+	latest "github.com/in-toto/in-toto-golang/in_toto/slsa_provenance/v0.1"
```
shibumi, line 18: To be honest, I am a little bit uncomfortable with "latest". This is a very generic key word. Can we rename this slsa?

shibumi: What do you think @adityasaky ? Or is it just me?

in_toto/slsa_provenance/v0.1/provenance.go
```diff
@@ -0,0 +1,63 @@
...
+// For in depth information refer to the specifictaion:
+// https://github.com/in-toto/attestation/blob/v0.1.0/spec/predicates/provenance.md
+type ProvenanceComplete struct {
+	Arguments   bool `json:"arguments"`
+	Environment bool `json:"environment"`
+	Materials   bool `json:"materials"`
+}
+
+/*
+DigestSet contains a set of digests. It is represented as a map from
```
shibumi, line 60: Can you change the comment style here? I know that we are not 100% consistent in our code, but I would like to move to single line comments only.

What do you think @adityasaky ?

shibumi:
Hi @priyawadhwa I left a few comments. Most of the code looks fine for me :) The comments are "discussable" and just reflect my own opinion. If you have good arguments I am easy to convince otherwise.

in_toto/model.go
adityasaky: Yeah, I think we can get more specific than `latest` to avoid any confusion / reuse in local blocks.

in_toto/slsa_provenance/v0.1/provenance.go
adityasaky: Sure. Worth noting that this piece was only moved over, but let's sneak it in. :smile:

adityasaky:
We're probably good to go on this one post @shibumi's comments, the `latest` one in particular.

in_toto/model.go
priyawadhwa: sgtm, renamed to `slsa` :)

in_toto/slsa_provenance/v0.1/provenance.go
priyawadhwa: fixed!

priyawadhwa:
Thanks @shibumi & @adityasaky , this should be ready for another look!

adityasaky (approved):

adityasaky:
Thanks, @priyawadhwa!

PR #175, opened 2022-06-30
Use lstrip when recording links for test-verify

adityasaky:
**Fixes issue:** N/A

**Description:**

`test-verify` in the Makefile generates link metadata but does not strip out the path of the temporary directory the artifact is stored in. This makes verification fail because the artifact rules don't expect the temporary directory.

requirements:**


SantiagoTorres:
LGTM

Issue #176, opened 2022-07-01
Support older SLSA version and v1.0


chuangw6:
In today's [in-toto community meeting](https://hackmd.io/@lukpueh/ry_e70Qqw#07012022-Meeting) (2022-07-01), we discussed that we want to have namespaced implemetation for the statement struct with [SLSA v1](https://github.com/slsa-framework/slsa-proposals/blob/main/0002/README.md) predicate when it comes out in next couple of months. This way, we can maintain the current [slsav0.2-specific statement struct](https://github.com/in-toto/in-toto-golang/blob/b4a6b184fd3ad4658e5018bc58e5f5a89fb23da5/in_toto/model.go#L986-L990) to allow a deprecation window.

(https://github.com/in-toto/attestation/issues/98 tracks all predicate types for the intoto statement)

If you have any questions or comments, feel free to add them here.

chuangw6:
cc @adityasaky @SantiagoTorres @joshuagl @mlieberman85

adityasaky:
Adding explicit models for v0.1 and v0.2 right now should be a minor diff, and will probably help us evaluate the work that goes into keeping multiple provenance versions around. Do you think you can open a PR to support v0.1 @chuangw6?

mlieberman85:
In the same vein does also make sense to separate out specific in-toto golang predicates that are supported to its own library? I don't want to conflate the two issues but I wonder if we might run into issues where we want to have a separate release cadence for in-toto-golang itself from community supported in-toto predicates.

chuangw6:
> Adding explicit models for v0.1 and v0.2 right now should be a minor diff, and will probably help us evaluate the work that goes into keeping multiple provenance versions around. Do you think you can open a PR to support v0.1 @chuangw6?

sgtm! Added #177 Please take a took.

chuangw6:
Hi @mlieberman85 ,
sorry for the delay.

> In the same vein does also make sense to separate out specific in-toto golang predicates that are supported to its own library? 

Do you mean each slsa version has its own repository in the `in-toto` project? 

> I don't want to conflate the two issues but I wonder if we might run into issues where we want to have a separate release cadence for in-toto-golang itself from community supported in-toto predicates.

Can you please explain a bit or provide a bit more context? 
Thank you!

adityasaky:
> I don't want to conflate the two issues but I wonder if we might run into issues where we want to have a separate release cadence for in-toto-golang itself from community supported in-toto predicates.

[...]

PR #177, opened 2022-07-13
Add slsa0.1 statement struct back

chuangw6:
**Description:**
As per https://github.com/in-toto/in-toto-golang/issues/176, we want to
maintain multiple versions of slsa-specific statement struct especially
when slsa v1 comes out in next couple of months. Reverting slsa0.1 back
helps the community evaluate the work.

Summary:
- Version specific statement structs are added (`ProvenanceStatementSLSA01`
and `ProvenanceStatementSLSA02`).
- The old `ProvenanceStatement` struct specific to slsa 0.2 is
still available, but will be officially deprecated in favour of
version explicit alternatives.
- Some common types shared between 0.1 and 0.2 are moved to the newly added
`common` package including `DigestSet`, `ProvenanceBuilder` and `ProvenanceMaterial`.
This might be a breaking change for the downstream users who use those
types to construct a statement.

If you have any questions/comments, please feel free to comment on this issue.

requirements:**

### DIFF
--- a/in_toto/model.go
+++ b/in_toto/model.go
@@ -15,7 +15,9 @@ import (
 	"strings"
 	"time"
 
-	slsa "github.com/in-toto/in-toto-golang/in_toto/slsa_provenance/v0.2"
+	"github.com/in-toto/in-toto-golang/in_toto/slsa_provenance/common"
+	slsa01 "github.com/in-toto/in-toto-golang/in_toto/slsa_provenance/v0.1"
+	slsa02 "github.com/in-toto/in-toto-golang/in_toto/slsa_provenance/v0.2"
 
 	"github.com/secure-systems-lab/go-securesystemslib/cjson"
 	"github.com/secure-systems-lab/go-securesystemslib/dsse"
@@ -964,8 +966,8 @@ func (mb *Metablock) Sign(key Key) error {
 
 // Subject describes the set of software artifacts the statement applies to.
 type Subject struct {
-	Name   string         `json:"name"`
-	Digest slsa.DigestSet `json:"digest"`
+	Name   string           `json:"name"`
+	Digest common.DigestSet `json:"digest"`
 }
 
 // StatementHeader defines the common fields for all statements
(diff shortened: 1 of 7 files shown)


in_toto/slsa_provenance/common/common.go
```diff
@@ -0,0 +1,5 @@
+package common
+
+// DigestSet contains a set of digests. It is represented as a map from
+// algorithm name to lowercase hex-encoded value.
+type DigestSet map[string]string
```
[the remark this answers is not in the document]
adityasaky: Yes, I like the look of this more than re-declaring in each version.

in_toto/model.go
```diff
@@ -985,10 +987,16 @@ type Statement struct {
 	Predicate interface{} `json:"predicate"`
 }
 
-// ProvenanceStatement is the definition for an entire provenance statement.
-type ProvenanceStatement struct {
```
adityasaky, line 989: Should we keep this around and deprecate it so that users switch to explicitly versioned types?

chuangw6: SG! Changed!

in_toto/slsa_provenance/common/common.go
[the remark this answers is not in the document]
chuangw6: I am now a little bit hesitate about moving these into a `common` package. I saw some other packages using the `DigestSet` type directly when creating `subjects` and `materials` i.e. [Tekton Chains](https://github.com/tektoncd/chains/blob/main/pkg/chains/formats/intotoite6/intotoite6.go#L139-L167). Moving these new types to a new package might break existing customers. If it's only breaking Tekton Chains, I am happy to coordinate to make Chains be compatible, but not sure other dependencies.

wdyt?

in_toto/model.go
chuangw6: SG! Should I add this comment after slsa v1 comes out or now?

shibumi: Adding `// Deprecated: ..` sounds good to me

in_toto/model.go
```diff
@@ -985,10 +987,16 @@ type Statement struct {
 	Predicate interface{} `json:"predicate"`
 }
 
-// ProvenanceStatement is the definition for an entire provenance statement.
+// ProvenanceStatement_SLSA01 is the definition for an entire provenance statement with SLSA 0.1 predicate.
+type ProvenanceStatement_SLSA01 struct {
```
shibumi, line 991: underscores are actually not recommended in Golang. I think I would go for ProvenanceStatementSLSA01 instead.

in_toto/slsa_provenance/common/common.go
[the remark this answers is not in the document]
chuangw6: Update: I've moved more common types to this common `package` including `ProvenanceBuilder` and `ProvenanceMaterial`.

adityasaky (approved):

chuangw6:
> Should we cut a release and make changes to consumers downstream off of that version?

Thanks @adityasaky for approving this PR and getting it merged! Cutting a new release SGTM

not sure the upgrade/deprecation policies of other consumers. For Tekton Chains, I am definitely happy to take care of it 😃

PR #185, opened 2022-10-13
Update Go versions used in CI

adityasaky:
Switch from testing Go `1.16.x`, `1.17.x` to `1.17.x`, `1.18.x`, `1.19.x`.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -4,7 +4,7 @@ jobs:
   test:
     strategy:
       matrix:
-        go-version: [1.16.x, 1.17.x]
+        go-version: [1.18.x, 1.19.x]
         os: [ubuntu-latest, macos-latest, windows-latest]
     runs-on: ${{ matrix.os }}
     steps:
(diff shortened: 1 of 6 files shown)


shibumi:
I have restarted the failing job. If it still fails, we might have to look into this.

shibumi:
> 1.17.x, 1.18.x, 1.19.x

Does it make sense to support 3 versions? Even Go supports only 2 LTS releases in a row.

shibumi:
That's weird. This line fails: `run: test -z $(go fmt ./...)`, but why is this succeeding on every other OS?

shibumi:
Can you manually run `go fmt ...` on a Linux system and see if code needs to be formatted?

pxp928:
@adityasaky Did some testing and seems like 1.19 formats the code a bit differently. Did a local test -> https://github.com/pxp928/in-toto-golang/pull/2 and it does need some files to be formatted.

adityasaky:
Yes, I just expected the error to be different but thinking about it some more, the output of fmt is probably too long as an argument to `test`.

shibumi:
> I thought we were following the previous, current, next model, but I'm okay with just doing two. 😄

But Isn't Go 1.19 already released? If we follow the 'next' model this would be a release candidate of Go 1.20. 🤔

adityasaky:
Good point. Do we switch to 1.18 and 1.19 then? And fmt only on latest?

shibumi:
> Good point. Do we switch to 1.18 and 1.19 then? And fmt only on latest?

Yes, I would suggest doing this. It sucks that they changed the format.. but we have to live with it.

shibumi (approved):
LGTM

Issue #187, opened 2022-10-17
CycloneDX predicate is wrong


lehors:
**General information:**

CycloneDX support was added in PR #169 by defining a predicate for it but its value is wrong.

**Description of the bug:**

The CycloneDX predicate is currently defined as "https://cyclonedx.org/schema" but the CycloneDX project has decided to use "https://cyclonedx.org/bom" instead. The latter is already used by some tools like syft and the discrepancy is creating incompatibilities with tools like cosign that are using the definition from this package.

**Anything special you want to tell us?**

For more info on this see: https://github.com/in-toto/attestation/issues/82#issuecomment-1280759933 and
https://github.com/in-toto/attestation/issues/82#issuecomment-1280983953

shibumi:
Fixed with: https://github.com/in-toto/in-toto-golang/pull/188#event-7614433625

PR #188, opened 2022-10-17
Update CycloneDX predicate

lehors:
**Fixes issue:**

This addresses Issue #187 and resolves an incompatibility between cosign and syft. See: https://github.com/anchore/syft/issues/1268

Signed-off-by: Arnaud J Le Hors <lehors@us.ibm.com>

**Description:**

This changes the value of the predicate to be in line with the decision from the CycloneDX project: https://github.com/in-toto/attestation/issues/82#issuecomment-1280983953 and what other tools like syft are already using.
This will make cosign and syft compatible.

### DIFF
--- a/in_toto/model.go
+++ b/in_toto/model.go
@@ -82,7 +82,7 @@ const (
 	// version.
 	PredicateSPDX = "https://spdx.dev/Document"
 	// PredicateCycloneDX represents a CycloneDX SBOM
-	PredicateCycloneDX = "https://cyclonedx.org/schema"
+	PredicateCycloneDX = "https://cyclonedx.org/bom"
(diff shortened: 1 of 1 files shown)


adityasaky:
cc @puerco I see you added support for cyclonedx originally and you should probably be aware of this change.

puerco:
Thanks for tagging me @adityasaky . LGTM. Once this merges I can bump the in-toto dependency on cosign. I just checked and sigstore prefers tagged versions but if no release is coming soon, we should be able to pull the go module @ HEAD.

shibumi (approved):

adityasaky:
I think we can cut a release soon @puerco, no need to pull from HEAD.

Issue #197, opened 2023-01-19
Support v1.0 SLSA provenance format


asraa:
**Description of the feature request:**

Support v1.0 SLSA provenance, as defined in a draft here: https://github.com/slsa-framework/slsa/pull/525

This will be merged soon, and so I can help to start a PR defining the struct, so long as the type is defined with a `?draft` URI for now. (Or we can wait to merge it).

One item to consider here that I've always wondered is  should we be using generated Go code against the protobuf message defining the predicate? This would mean that the oneof support comes out of the box. On the other hand, there's some overhead of bringing in the proto runtime dependencies and complexity.

 The oneof could also be handled by an interface, but this is also a little bit of overhead.

cc @marklodato @rbehjati @laurentsimon

PR #200, opened 2023-02-06
feat: add v1?draft slsa provenance definition

asraa:
Signed-off-by: Asra Ali <asraa@google.com>

**Fixes issue:**
Related https://github.com/in-toto/in-toto-golang/issues/197

**Description:**
Implements the `v1?draft` SLSA provenance spec. I expect it will be complete by end of the month, and don't think there will be too many changes. I can keep the issue open to migrate the URI to `v1` (removing the `draft`) with whatever remaining format changes there are in the follow-up.

requirements:**

- [ x ] Tests have been added for the bug fix or new feature
- [ x ] Docs have been added for the bug fix or new feature: added docstrings

### DIFF
--- a/in_toto/slsa_provenance/v1.0/provenance.go
+++ b/in_toto/slsa_provenance/v1.0/provenance.go
@@ -1,0 +1,141 @@
+package v1
+
+import (
+	"time"
+
+	"github.com/in-toto/in-toto-golang/in_toto/slsa_provenance/common"
+)
+
+const (
+	// PredicateSLSAProvenance represents a build provenance for an artifact.
+	PredicateSLSAProvenance = "https://slsa.dev/provenance/v1?draft"
+)
+
+// ProvenancePredicate is the provenance predicate definition.
+type ProvenancePredicate struct {
+	// The BuildDefinition describes all of the inputs to the build. The
+	// accuracy and completeness are implied by runDetails.builder.id.
+	//
+	// It SHOULD contain all the information necessary and sufficient to
+	// initialize the build and begin execution.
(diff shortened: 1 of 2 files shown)


pxp928 (approved):
LGTM! We can keep the issue open and change it to remove `draft` once finalized.

adityasaky:
I'd prefer to leave it open as a draft to avoid someone accidentally using a non tagged version. @asraa wdyt?

asraa:
> I'd prefer to leave it open as a draft to avoid someone accidentally using a non tagged version. @asraa wdyt?

Right now the format is in `v1?draft` URI, so I do think people will be aware that the provenance is subject to change.

The reason I'd like it (even if pushed to a branch?) is that it blocks us from drafting tooling until the spec goes GA, where it would be nice to be developing alpha tooling while it still is to test its use-ability. WDYT about the branch idea?

adityasaky:
Gotcha, let me review what's the status of `next`.

pxp928:
> > I'd prefer to leave it open as a draft to avoid someone accidentally using a non tagged version. @asraa wdyt?
> 
> Right now the format is in `v1?draft` URI, so I do think people will be aware that the provenance is subject to change.
> 
> The reason I'd like it (even if pushed to a branch?) is that it blocks us from drafting tooling until the spec goes GA, where it would be nice to be developing alpha tooling while it still is to test its use-ability. WDYT about the branch idea?

I agree. Would be useful to start updating tools to meet the spec.

adityasaky:
Done! Thanks @asraa!

asraa:
Thank you all! I appreciate having a basis :)
I will definitely be on top of any spec updates, so that's my contract for this.

PR #221, opened 2023-04-14
feat: Finalize the SLSA v1.0 provenance format

asraa:
**Fixes issue:**
Fixes https://github.com/in-toto/in-toto-golang/issues/197

**Description:**
Finalizes the SLSA v1.0 provenance format.

@adityasaky it would be great to merge the `next` branch into main soon and do a minor release with the SLSA v1 feature!

requirements:**

- [ x ] Tests have been added for the bug fix or new feature

cc @marklodato @laurentsimon @chuangw6

### DIFF
--- a/in_toto/slsa_provenance/v1.0/provenance.go
+++ b/in_toto/slsa_provenance/v1.0/provenance.go
@@ -8,7 +8,7 @@ import (
 
 const (
 	// PredicateSLSAProvenance represents a build provenance for an artifact.
-	PredicateSLSAProvenance = "https://slsa.dev/provenance/v1?draft"
+	PredicateSLSAProvenance = "https://slsa.dev/provenance/v1"
 )
 
 // ProvenancePredicate is the provenance predicate definition.
@@ -48,20 +48,20 @@ type ProvenanceBuildDefinition struct {
 	// becomes.
 	ExternalParameters interface{} `json:"externalParameters"`
 
-	// The parameters that are under the control of the builder. The primary
-	// intention of this field is for debugging, incident response, and
-	// vulnerability management. The values here MAY be necessary for
-	// reproducing the build. There is no need to verify these parameters
-	// because the build system is already trusted, and in many cases it is not
-	// practical to do so.
-	SystemParameters interface{} `json:"systemParameters,omitempty"`
+	// The parameters that are under the control of the entity represented by
+	// builder.id. The primary intention of this field is for debugging,
+	// incident response, and vulnerability management. The values here MAY be
+	// necessary for reproducing the build. There is no need to verify these
+	// parameters because the build system is already trusted, and in many cases
+	// it is not practical to do so.
+	InternalParameters interface{} `json:"internalParameters,omitempty"`
 
 	// Unordered collection of artifacts needed at build time. Completeness is
 	// best effort, at least through SLSA Build L3. For example, if the build
 	// script fetches and executes “example.com/foo.sh”, which in turn fetches
 	// “example.com/bar.tar.gz”, then both “foo.sh” and “bar.tar.gz” SHOULD be
 	// listed here.
-	ResolvedDependencies []ArtifactReference `json:"resolvedDependencies,omitempty"`
+	ResolvedDependencies []ResourceDescriptor `json:"resolvedDependencies,omitempty"`
 }
 
 // ProvenanceRunDetails includes details specific to a particular execution of a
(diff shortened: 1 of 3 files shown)


chuangw6 (commented):
Thank you @asraa !!
Can we add a provenance statement struct in model.go like [these](https://github.com/in-toto/in-toto-golang/blob/a813798404c5f3d9ed27ec14ef4915f7e913b375/in_toto/model.go#L993-L1010)?

chuangw6:
Also want to loop in @sherzberg-1 as fyi. Just want to make sure the intoto provenance statement can be serialized and deserialized to grafeas proto.

sherzberg-1:
Thanks @chuangw6 , will make sure they can serialize/deserialize back and forth.

MarkLodato:
LGTM

MarkLodato (approved):

chuangw6:
Hi @adityasaky ,
[SLSA v1 is released](https://openssf.org/press-release/2023/04/19/openssf-announces-slsa-version-1-0-release/). And I think this PR is up to date.
Can you please TAL at this PR and see if we can get the branch into main and cut a release? 
Thanks!

in_toto/slsa_provenance/v1.0/provenance.go
```diff
@@ -86,29 +86,39 @@ type ProvenanaceRunDetails struct {
...
-// ArtifactReference describes a particular artifact. At least one of URI or
-// digest MUST be specified.
-type ArtifactReference struct {
-	// URI describing where this artifact came from. When possible, this SHOULD
-	// be a universal and stable identifier, such as a source location or
-	// Package URL (purl).
+// ResourceDescriptor describes a particular software artifact or resource
+// (mutable or immutable).
+// See https://github.com/in-toto/attestation/blob/main/spec/v1.0/resource_descriptor.md
+type ResourceDescriptor struct {
```
adityasaky, line 95: Thoughts on using ResourceDescriptor from https://github.com/in-toto/attestation/blob/main/go/v1/resource_descriptor.pb.go?

cc @marcelamelara

adityasaky (commented):
Thanks @asraa! I think we can merge this overall.

However, I do want to hear from @in-toto/attestation-maintainers re using the models generated for ResourceDescriptor (and possibly Statement etc as well).

in_toto/slsa_provenance/v1.0/provenance.go
chuangw6: I think that's a good idea. But given the proto only defines ResourceDescriptor and misses many other fields of SLSA v1 predicate, should we move on with current approach and cut a release first, and later switch over to models generated from proto once all v1 predicate fields are well defined in attestation repo?

marcelamelara: To my understanding, SLSA already defines full protos for SLSA provenance v1: https://github.com/slsa-framework/slsa/tree/main/docs/provenance/schema/v1

We (in-toto attestation maintainers) were operating under the assumption that SLSA will maintain all SLSA-related predicate spec and proto versions. Based on your comment here, it sounds like there's an expectation that the in-toto attestation repo should maintain (a separate copy) of SLSA provenance v1 protos?

chuangw6: Oh, I was only looking at https://github.com/in-toto/attestation/tree/main/protos/in_toto_attestation/v1. I thought intoto will define its own proto again.

If intoto can reuse the proto from SLSA, that would be great. I have no option on this :D I'd defer intoto maintainers to decide this.

I am looking forward to seeing an intoto-golang release for SLSA v1!

marcelamelara: Got it :) At least my preference would be to use the SLSA-defined protos for generating the language bindings for the predicate, so this integration is something we should definitely discuss on our end.

chuangw6: Sounds good to me! Is there a target date we can expect a release with the slsa v1 support? Thanks!

in_toto/slsa_provenance/v1.0/provenance.go
```diff
@@ -8,7 +8,7 @@ import (
 
 const (
 	// PredicateSLSAProvenance represents a build provenance for an artifact.
-	PredicateSLSAProvenance = "https://slsa.dev/provenance/v1?draft"
+	PredicateSLSAProvenance = "https://slsa.dev/provenance/v1"
```
adityasaky, line 11: Since the predicate type will not include the minor version, can this package dir be just `v1` instead of `v1.0`?

> Minor version changes are always backwards compatible and “monotonic.” Such changes do not update the predicateType.

https://slsa.dev/provenance/v1#parsing-rules

in_toto/slsa_provenance/v1.0/provenance.go
adityasaky: I think I'm okay to merge this PR and then next into main and cut a release to unblock other tools. My inclination is to switch it out behind the scenes later as we want to do that for multiple models, not just SLSA Provenance.

adityasaky (approved):
LGTM, I do have a question about the package directory.

in_toto/slsa_provenance/v1.0/provenance.go
pxp928: +1 with @adityasaky can merge for now and switch behind the scenes in another release.

pxp928 (approved):
LGTM

in_toto/slsa_provenance/v1.0/provenance.go
chuangw6: > My thinking is that if Provenance v1.1 introduces a change, it'd go in in_toto/slsa_provenance/v1/provenance.go. I'm wary of having in_toto/slsa_provenance/v1.0/provenance.go, in_toto/slsa_provenance/v1.1/provenance.go, etc. Since they're backwards compatible from here, I see no reason why we can't have a single v1/provenance.go.

SGTM and agreed!

PR #222, opened 2023-04-20
Various fixes

### LINKED ISSUE #211: Verification using ED25519 keys for functionaries returns error
**General information:**

I am using a similar setup as described in #207.
In the demo project I replaced the keys of bob and carl by ED25519 versions in PEM format. I created the keys in the  `root.layout` using the command `in-toto key layout bob/carl`. 

* Version (tag or commit): checked out repo at  state of commit 3dcfea315524dd5eddd44ee0959b971251cce0bd
* Operating system (Linux/Mac/Windows): Debian 11.6 x86 architecture
* Go version: `go version go1.20.1`
* Go build flags (if any):

**Description of the bug:**

Running all the steps in demo-project works without an issue unless I want to verify the result. When running the verification I get the following: 

```
in-toto verify --layout root.layout --layout-keys alice.pub

inspection failed: invalid signature: crypto/rsa: verification error
```

Yet when creating everything up until the verification step using the Golang verison of in-toto and then at the verification step I use the python version instead the verification is successful. The command I ran in the end is: 

```
in-toto-verify --layout root.layout --layout-key alice.pub
```

The exit code is zero as it would be expected.

Due to the error message I am assuming that instead of using ED25519 for the verification of the corresponding signatures, RSA is used.

adityasaky:
#211: This PR addresses the inclusion of empty `private` when creating and verifying signatures. @anotherbridge I wonder if you could build off this branch and see if the issue is fixed as well? :smile: 

The PR also includes minor clean up of imports, type assertions etc.


adityasaky:
Hm, windows is strange :thinking:

anotherbridge:
Hey, @adityasaky I just build and tested the in-toto version from the branch you created the PR. I did exactly the same steps as in #211. The layout generated was still the same, i.e. the private field was missing and still the `in-toto verfiy` run results in: 

```
inspection failed: invalid signature: crypto/rsa: verification error
```

in_toto/keylib.go
```diff
@@ -384,9 +384,9 @@ func getDefaultKeyScheme(key interface{}) (scheme string, keyIDHashAlgorithms []
 
 func (k *Key) loadKey(key interface{}, pemData *pem.Block, scheme string, keyIDHashAlgorithms []string) error {
 
-	switch key.(type) {
+	switch K := key.(type) {
```
pxp928, line 387: this does not need to be capital "K". No need to make it exported. Can we use another variable name if it conflicts with the function.

pxp928 (approved):
LGTM. One comment

adityasaky:
Also cleaned up dependabot a bit @pxp928.

pxp928 (approved):

PR #151, opened 2021-10-15
feat: use-dsse flag add to run command

developer-guy:
Signed-off-by: Batuhan Apaydın <batuhan.apaydin@trendyol.com>

**Fixes issue:**

Fixes #148 

**Description:**

If I did something wrong, I want to apologize in advance, but I wanted to do something, at least I wanted to start somewhere 🧑🏻‍💻🔮

**Please verify and check that the pull request fulfills the following
requirements:**

- [ ] Tests have been added for the bug fix or new feature
- [ ] Docs have been added for the bug fix or new feature

Issue #148, opened 2021-10-06
Allow generating in-toto links using DSSE


adityasaky:
**Description of the feature request:**

in-toto-golang should include a user-specified option that controls if the link metadata generated uses the current legacy signature wrapper or DSSE. The library should also be able to load these links and verify them. in-toto can use https://github.com/secure-systems-lab/go-securesystemslib for the DSSE provider.

adityasaky:
It'd have to use both. We'd be using the signing parts in `InTotoRun` to generate the link metadata using DSSE, and the verification part in `InTotoVerify`.

adityasaky:
(I should add that it would be fantastic if you could help out on this front!)

PR #228, opened 2023-05-01
Make in-toto-golang sig env agnostic, enable DSSE

adityasaky:
**Fixes issue:** #148, supersedes #151 

**Description:**

Makes in-toto-golang agnostic to the signature wrapper used via a generic `Metadata` interface that can handle either Metablock or DSSE.

requirements:**


adityasaky:
So far, it doesn't implement DSSE support, it just introduces support via the Metadata interface. I'm waiting on a couple of PRs in go-securesystemslib to move to use here.

adityasaky:
I've added some tests to runlib for the DSSE envelope. I'm missing more comprehensive tests for verifylib. I have a simple one using a DSSE layout but the more complicated one is going to be regenerating link metadata using DSSE and storing them in a non confusing way. The best bet may be to rename the steps `write-code -> write-code-dsse` and update the artifact rules as well. WDYT @pxp928?

adityasaky:
BTW, this PR is blocked on an upstream release of go-securesystemslib.

cmd/run.go
```diff
@@ -167,15 +174,15 @@ func run(cmd *cobra.Command, args []string) error {
...
 	}
 
-	block, err := intoto.InTotoRun(stepName, runDir, materialsPaths, productsPaths, args, key, []string{"sha256"}, exclude, lStripPaths, lineNormalization, followSymlinkDirs)
+	metadata, err := intoto.InTotoRun(stepName, runDir, materialsPaths, productsPaths, args, key, []string{"sha256"}, exclude, lStripPaths, lineNormalization, followSymlinkDirs, useDSSE)
 	if err != nil {
 		return fmt.Errorf("failed to create link metadata: %w", err)
 	}
 
-	linkName := fmt.Sprintf(intoto.LinkNameFormat, block.Signed.(intoto.Link).Name, key.KeyID)
+	linkName := fmt.Sprintf(intoto.LinkNameFormat, metadata.GetPayload().(intoto.Link).Name, key.KeyID)
```
pxp928, line 182: `metadata.GetPayload().(intoto.Link)` add check for this type assertion to not cause a panic

in_toto/envelope.go
```diff
@@ -0,0 +1,197 @@
...
+	if err != nil {
+		return err
+	}
+
+	ev, err := dsse.NewEnvelopeVerifier(verifier)
+	if err != nil {
+		return err
+	}
+
+	_, err = ev.Verify(context.Background(), e.envelope)
```
pxp928, line 105: missing check for err?

in_toto/envelope.go
```diff
@@ -0,0 +1,197 @@
...
+
+		var payload map[string]any
+		if err := json.Unmarshal(payloadBytes, &payload); err != nil {
+			panic(fmt.Errorf("error decoding payload: %w", err))
+		}
+
+		if payload["_type"] == "link" {
+			var link Link
+			if err := checkRequiredJSONFields(payload, reflect.TypeOf(link)); err != nil {
+				panic(fmt.Errorf("error decoding payload: %w", err))
```
pxp928, line 65: could you return the error and handle it upstream instead? Having a Panic here doesn't feel right.

in_toto/examples_test.go
```diff
@@ -33,11 +33,11 @@ func ExampleInTotoVerify() {
...
 	// Perform in-toto software supply chain verification, using the provided
 	// test data.
-	var layoutMb Metablock
-	if err := layoutMb.Load(LayoutPath); err != nil {
+	layoutMb, err := LoadMetadata(LayoutPath)
+	if err != nil {
 		fmt.Printf("Unable to load layout metadata: %s", err)
 	}
-	if err := validateLayout(layoutMb.Signed.(Layout)); err != nil {
+	if err := validateLayout(layoutMb.GetPayload().(Layout)); err != nil {
```
pxp928, line 40: type assertion check to not cause a panic.

in_toto/model.go
```diff
@@ -704,6 +677,105 @@ func validateLayout(layout Layout) error {
 	return nil
 }
 
+type Metadata interface {
+	Sign(Key) error
+	VerifySignature(Key) error
+	GetPayload() any
```
pxp928, line 683: not sure you can change the interface to return an error also.

in_toto/verifylib.go
```diff
@@ -93,9 +93,9 @@ func verifyMatchRule(ruleData map[string]string,
 	var dstArtifacts map[string]interface{}
 	switch ruleData["dstType"] {
 	case "materials":
-		dstArtifacts = dstLinkMb.Signed.(Link).Materials
+		dstArtifacts = dstLinkEnv.GetPayload().(Link).Materials
```
pxp928, line 96: check type?

pxp928:
> I've added some tests to runlib for the DSSE envelope. I'm missing more comprehensive tests for verifylib. I have a simple one using a DSSE layout but the more complicated one is going to be regenerating link metadata using DSSE and storing them in a non confusing way. The best bet may be to rename the steps write-code -> write-code-dsse and update the artifact rules as well. WDYT @pxp928?

Yup, that makes sense.

in_toto/envelope.go
pxp928: I see what you are doing. Nvm

cmd/run.go
adityasaky: InTotoRun can only return a link, should we add the check nonetheless?

in_toto/envelope.go
adityasaky: I'm not sure if `Do()` can pass that up, I agree about panic. Should we rework this to get away with sync.Once?

in_toto/examples_test.go
adityasaky: So once again we're feeding in the data here, demo.layout.

I added a check in InTotoVerify to ensure it's a Layout more generally.

in_toto/model.go
adityasaky: I think I'd prefer to rework how the required field validations happens to avoid using sync.Once and retain GetPayload without return an error.

in_toto/verifylib.go
adityasaky: We check to ensure it's not a layout prior to this in InTotoVerify. I guess a general question is if these checks need to happen everywhere for resilience?

adityasaky:
I added a DSSE stack of tests @pxp928 as well.

cmd/run.go
pxp928: I am ok with skipping the check if it will always be one type. Just in case it could return other types in the future, we would have to add the check.

in_toto/envelope.go
pxp928: hmmm....I think that would be cleaner. How much work would that take?

in_toto/verifylib.go
pxp928: Mostly for resilience as it will cause a panic

in_toto/envelope.go
adityasaky: let me see.

cmd/run.go
adityasaky: Agreed but we'd have to handle other things at the time anyway including where to get the `Name` etc. I'd prefer to handle this separately when the time comes. I think we may have a world where `InTotoRun` generates other supported predicates.

in_toto/verifylib.go
adityasaky: #230

in_toto/envelope.go
adityasaky: https://github.com/in-toto/in-toto-golang/pull/228/commits/71d35ad4a95b009588d473f396bdf0bdc0a463b1 has a tentative patch for this.

pxp928 (approved):
LGTM!

PR #238, opened 2023-06-01
Drop use of `any` for hash objects

adityasaky:
**Fixes issue:**
N/A

**Description:**
We know that hash objects are a map from string to string. We can therefore set that expectation directly rather than using interface types.

requirements:**

### DIFF
--- a/in_toto/attestations_test.go
+++ b/in_toto/attestations_test.go
@@ -429,10 +429,10 @@ func TestLinkStatement(t *testing.T) {
     "name": "name",
     "command": ["cc", "-o", "baz", "baz.z"],
     "materials": {
-       "kv": "vv"
+       "kv": {"alg": "vv"}
     },
     "products": {
-       "kp": "vp"
+       "kp": {"alg": "vp"}
     },
     "byproducts": {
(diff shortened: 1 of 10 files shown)


adityasaky:
@shibumi any thoughts on if the hash obj needs to be an interface?

shibumi:
Hey,

first of all +1 for replacing all `interface{}` with `any`. That's the recommended way to go for the future.

Was there any reason why we chose interfaces at the first place ? 🤔  If we are 100% sure that we are dealing with `map[string]string` objects here, there is no point in using interfaces. Using the correct types would make everything tremendously easier.

I think the idea behind using interfaces was that we might insert other data types than `string` in there in the future.

adityasaky:
I'm not so sure if any dependencies rely on interfaces. Most dependencies don't use links afaik so this change may be fine with warnings. Alternatively, since we're transitioning to link predicates which use resource descriptors, we can close this. We'll deprecate legacy links at some point anyway.

shibumi (approved):

adityasaky:
Rebased to resolve conflicts.

Issue #253, opened 2023-07-17
ALLOW pattern doesn't work as expected on windows


PradyumnaKrishna:
**General information:**

* Version: latest (0.9.0)
* Operating system: Windows
* Go version: 1.20

**Description of the bug:**
On testing in-toto-golang in windows and running demo against it, I found that in-toto-golang doesn't ALLOWs or ignores subdirs on windows. On running in-toto/demo, the in-toto-golang verify commands doesn't `ALLOW demo-project/.git/*` as provided in layout and fails verification with message disallowed by rule [DISALLOW *].

The logs https://github.com/PradyumnaKrishna/in-toto-testing/actions/runs/5572000137/jobs/10177593630#step:7

```
======================================================================
FAIL: test_using_demo (tests.test_golang.TestGolang)
Test running in-toto demo against in-toto implementations.
----------------------------------------------------------------------
Traceback (most recent call last):
  File "D:\a\in-toto-testing\in-toto-testing\tests\test_golang.py", line 146, in test_using_demo
    self.assert_cli_sys_exit(verify_cmd_go, 0)
  File "D:\a\in-toto-testing\in-toto-testing\tests\common.py", line 48, in assert_cli_sys_exit
    self.assertEqual(exc.returncode, status, exc.stderr)
AssertionError: 1 != 0 : inspection failed: artifact verification failed for Inspection 'untar', products [
demo-project\.git\packed-refs
demo-project\.git\hooks\fsmonitor-watchman.sample
demo-project\.git\hooks\pre-applypatch.sample
demo-project\.git\logs\refs\remotes\origin\HEAD
demo-project\.git\objects\pack\pack-2c9aff0827d356448f801394c227db42ff9bedb6.pack demo-project\.git\info\exclude
demo-project\.git\description demo-project\.git\hooks\pre-rebase.sample
demo-project\.git\refs\heads\master demo-project\.git\hooks\sendemail-validate.sample
demo-project\.git\hooks\push-to-checkout.sample
demo-project\.git\objects\pack\pack-2c9aff0827d356448f801394c227db42ff9bedb6.rev
demo-project\.git\config demo-project\.git\logs\refs\heads\master
demo-project\.git\hooks\pre-merge-commit.sample
demo-project\.git\hooks\pre-push.sample
demo-project\.git\hooks\post-update.sample
demo-project\.git\objects\pack\pack-2c9aff0827d356448f801394c227db42ff9bedb6.idx
[...]

PR #254, opened 2023-07-18
Fixes filepath pattern matching in windows

PradyumnaKrishna:
Fixes: #253 

**Description:**
Filepath works differently compared to linux, on testing found it didn't allowed the provided path pattern in windows. This is fixed using `filepath.Match` instead of `match`.

requirements:**

### DIFF
--- a/in_toto/runlib.go
+++ b/in_toto/runlib.go
@@ -95,18 +95,19 @@ return value is the error.
 func RecordArtifacts(paths []string, hashAlgorithms []string, gitignorePatterns []string, lStripPaths []string, lineNormalization bool, followSymlinkDirs bool) (evalArtifacts map[string]interface{}, err error) {
 	// Make sure to initialize a fresh hashset for every RecordArtifacts call
 	visitedSymlinks = NewSet()
-	evalArtifacts, err = recordArtifacts(paths, hashAlgorithms, gitignorePatterns, lStripPaths, lineNormalization, followSymlinkDirs)
-	// normalize windows paths to unix paths
-	for key, value := range evalArtifacts {
-		delete(evalArtifacts, key)
-		key = filepath.ToSlash(key)
-
-		// Convert windows filepath to unix filepath.
-		evalArtifacts[key] = value
+	evalArtifactsUnnormalized, err := recordArtifacts(paths, hashAlgorithms, gitignorePatterns, lStripPaths, lineNormalization, followSymlinkDirs)
+	if err != nil {
+		return nil, err
 	}
 
-	// pass result and error through
-	return evalArtifacts, err
+	// Normalize all paths in evalArtifactsUnnormalized.
+	evalArtifacts = make(map[string]interface{})
+	for key, value := range evalArtifactsUnnormalized {
+		// Convert windows filepath to unix filepath.
+		evalArtifacts[filepath.ToSlash(key)] = value
+	}
+
+	return evalArtifacts, nil
 }
 
 /*
--- a/in_toto/runlib.go
+++ b/in_toto/runlib.go
@@ -101,7 +101,7 @@ func RecordArtifacts(paths []string, hashAlgorithms []string, gitignorePatterns 
 	}
 
 	// Normalize all paths in evalArtifactsUnnormalized.
-	evalArtifacts = make(map[string]interface{})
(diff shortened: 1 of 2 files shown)


PradyumnaKrishna:
I don't why `path.Match` doesn't matches for `*` pattern. I have used this hack around to fix the problem, but I guess this is not an ideal solution to this problem. There may be some other bugs around.

adityasaky:
I think this isn't the solution we want to go with. To recap, the rule you used is `ALLOW demo-project/.git/*` and the issue is that files like `demo-project\.git\packed-refs` are not matched. The issue isn't with `*` but rather that the pattern uses `/` and the file paths use `\`. Somewhere, the filepath in `demo-project\.git\packed-refs` should have gotten normalized so that it uses `/` as well, before the rules are applied.

shibumi:
I agree with @adityasaky here. The issue seems to be that the path did not get normalized.
in-toto-golang uses https://github.com/shibumi/go-pathspec. If you find a bug, we should probably fix it in go-pathspec instead.

The normalization does not seem to be the problem, because I fixed windows filepath normalization in: https://github.com/shibumi/go-pathspec/commit/36546338c10c7357baf58c374be304f1bde2732c

Do we use the correct version of go-pathspec?

@PradyumnaKrishna can you provide more information about what version of in-toto-golang you are using?

PradyumnaKrishna:
> @PradyumnaKrishna can you provide more information about what version of in-toto-golang you are using?

Thank you @shibumi for your explanation. `go-pathspec` wasn't used to get `artifactsPath` but to ignore or exclude artifacts present in the directory. I created a fix for it now I believe.

adityasaky:
Hmm, I'll take a look tomorrow. I'm still not sure why this normalization isn't affecting a) other rules like CREATE, MODIFY, and b) why previous tests didn't catch this.

PradyumnaKrishna:
> a) other rules like CREATE, MODIFY

I guess its affecting all rules but we didn't test them.

> b) why previous tests didn't catch this.

because we only ran tests with unix filepaths and not windows filepaths. I have added new test case for windows using windows filepaths.

adityasaky:
>because we only ran tests with unix filepaths and not windows filepaths

Ah I see, then I suppose we'd start with adding tests for the match.go file rather than verifylib and build from there?

adityasaky:
aaactually, thinking about it some more, you're right. We want match to be consistent and pass in normalized paths.

PradyumnaKrishna:
Does this PR needs any changes?

shibumi:
> > a) other rules like CREATE, MODIFY
> 
> I guess its affecting all rules but we didn't test them.
> 
> > b) why previous tests didn't catch this.
> 
> because we only ran tests with unix filepaths and not windows filepaths. I have added new test case for windows using windows filepaths.

Hey,

Sorry for the late reply, I was in surgery two weeks ago.

What is the current status and where are these test cases you are speaking about? I can't see any test cases in your PR.

PradyumnaKrishna:
> What is the current status and where are these test cases you are speaking about? I can't see any test cases in your PR.

Previously I normalized path in `verifyArtifacts` now I am normalizing paths in `recordArtifacts`. I have reset the head, if you want to browse previous code here it is https://github.com/PradyumnaKrishna/in-toto-golang/tree/605cfc53bed9397898e3b4d298147a77b7ea5d9f.

in_toto/runlib.go
```diff
@@ -96,6 +96,15 @@ func RecordArtifacts(paths []string, hashAlgorithms []string, gitignorePatterns
 	// Make sure to initialize a fresh hashset for every RecordArtifacts call
 	visitedSymlinks = NewSet()
 	evalArtifacts, err = recordArtifacts(paths, hashAlgorithms, gitignorePatterns, lStripPaths, lineNormalization, followSymlinkDirs)
+	// normalize windows paths to unix paths
+	for key, value := range evalArtifacts {
+		delete(evalArtifacts, key)
```
adityasaky, line 101: Modifying a map during iteration can sometimes lead to unpredictable behavior (though probably more when we're adding new keys). I recommend updating this to `evalArtifactsUnnormalized` or something and then creating a new `evalArtifacts` with the path normalized for each element in the former.

adityasaky (approved):
Thanks!

shibumi (approved):

shibumi:
Thanks, LGTM

Issue #260, opened 2023-08-24
Oversight in the SLAS v1 field `invocationID`?


chuangw6:
**Description of the bug:**

In the SLSA v1 provenance, the json tag for the field invocation id should be `invocationId` instead of `invocationID` per [slsa v1 spec](https://slsa.dev/spec/v1.0/provenance#schema) on the official website and [the proto](https://github.com/in-toto/attestation/blob/main/protos/in_toto_attestation/predicates/provenance/v1/provenance.proto).

But the tag in in-toto-golang is `invocationID` https://github.com/in-toto/in-toto-golang/blob/5fd5c6748041bbaf5fd05da2a16dcb2c96dc30b6/in_toto/slsa_provenance/v1/provenance.go#L144

chuangw6:
cc @adityasaky

adityasaky:
I think it's preferable to fix this by directly using the proto. I know @marcelamelara was interested in getting this done too, unfortunately I've been a little swamped. @marcelamelara do you have any cycles to help out here?

chuangw6:
> I think it's preferable to fix this by directly using the proto.

Agreed.

marcelamelara:
+1 to switching to using the protobufs in this implementation. IIRC, the last time I looked at making this change happen, it wasn't quite as simple as replacing one struct for the other, but I'll have _some_ cycles in the next week or two, so I can take a look.

msuozzo:
The aforementioned change is still not submitted so the proto version is effectively unusable. Can we merge the fix for this issue in the meantime? https://github.com/in-toto/in-toto-golang/pull/328

msuozzo:
Is this worth a patch-release? This is technically a bug fix but I can pin to the commit hash if necessary.

adityasaky:
I'll try and tag a release tomorrow. There are a couple pending updates etc we should probably merge as well, I'll check if someone has a few to take a pass at fixing some of the CI issues!

PR #267, opened 2023-09-19
Deprecate Provenance v1 struct in favor of /attestation protobufs

### LINKED ISSUE #265: Define intoto v1.0 statement type
As discussed in https://github.com/slsa-framework/slsa/issues/918, SLSA v1 predicate is supposed to be wrapped in [intoto v1.0 statement](https://github.com/in-toto/attestation/blob/main/spec/v1/statement.md).

Therefore, I suppose there should be a const defined for intoto v1.0 statement type i.e. `StatementInTotoV10`, but there appears to be none.

https://github.com/in-toto/in-toto-golang/blob/8a5dc9e6d8637cf72c4b5103216ce3445e053a14/in_toto/attestations.go#L10-L22

marcelamelara:
This PR is the first step towards making in-toto-golang ITE-9 compatible. It deprecates the SLSA Provenance v1 and ITE-6 structs defined in this package, which have fallen out of spec and been superseded by the pre-generated Go bindings based on the [proto definitions](https://github.com/in-toto/attestation/tree/main/protos) in the ITE-6/ITE-9 [attestation repo](https://github.com/in-toto/attestation).

Fixes #265 and #260 .

### DIFF
--- a/in_toto/attestations.go
+++ b/in_toto/attestations.go
@@ -22,12 +22,28 @@ const (
 )
 
 // Subject describes the set of software artifacts the statement applies to.
+//
+// Deprecated: This implementation of Subject exists for historical
+// compatibility and should not be used. This implementation has been
+// superseded by a ResourceDescriptor struct generated from the Protobuf
+// definition in
+// https://github.com/in-toto/attestation/tree/main/protos/in_toto_attestation/v1.
+// To generate an ITE-6 v1 Statement subject, use the ResourceDescriptor Go
+// APIs provided in https://github.com/in-toto/attestation/tree/main/go/v1.
 type Subject struct {
 	Name   string           `json:"name"`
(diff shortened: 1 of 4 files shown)


in_toto/attestations.go
```diff
@@ -1,6 +1,7 @@
 package in_toto
 
 import (
+	ita1 "github.com/in-toto/attestation/go/v1"
```
adityasaky, line 4: A `go mod tidy` is in order to add this to go.mod and go.sum.

adityasaky (approved):

in_toto/attestations.go
marcelamelara: Yes! I need to run `fmt` and all that on this, too.

pxp928 (approved):

PR #268, opened 2023-09-20
Add APIs for Statement v1 and SLSA Provenance v1 protos

### LINKED ISSUE #260: Oversight in the SLAS v1 field `invocationID`?

### LINKED ISSUE #265: Define intoto v1.0 statement type

marcelamelara:
This PR adds APIs for generating ITE-6 v1 Statements using the protobuf-based language bindings provided by the [in-toto Attestation Framework](https://github.com/in-toto/attestation/tree/main/go). This PR also adds protobuf-based APIs that replace the [now-deprecated](#267) SLSA Provenance v1 structs.

Fixes #260, fixes #265.


adityasaky:
I like the ideas here, is the plan to take this all the way through signing the statement using dsse as well?

marcelamelara:
> is the plan to take this all the way through signing the statement using dsse as well?

@adityasaky Ideally yes. The idea would be to have these sort of generators for each supported predicate type, and switch on the predicate type in a place like runlib before we sign the Statement and generate the DSSE. From the CLI perspective, we may introduce a predicate type flag. Other implementers, like the SLSA provenance generator for GHA might use these APIs directly.

What I don't have a solution for yet is how the the predicate contents themselves will be generated/passed into in-toto, especially in the case of informational ITE-9 predicates.

marcelamelara:
In the interest of keeping PRs small enough and digestible, I'm going to scope this PR at only introducing the new v1 APIs, but not integrating them into the greater `in-toto-golang CLI` tool. With #288 in-flight as well, I think we have what we need to add support for both v0.3 Links, v1 Provenance and other predicates to in-toto-golang.

in_toto/attestation/common/common.go
```diff
@@ -0,0 +1,29 @@
...
+)
+
+/*
+PredicatePbToStruct converts a given protobuf representation of
+an in-toto attestation Predicate into a generic Struct struct
+needed to construct an in-toto Attestation Framework v1
+compliant Statement. If there are any marshalling problems,
+this function returns an error.
+*/
+func PredicatePbToStruct(predicatePb proto.Message) (*structpb.Struct, error) {
```
adityasaky, line 16: I wonder if want this to be `internal/`, I don't think we want to accidentally have people use this. Or it could be a private function in `attestation`.

in_toto/attestation/v1/resource_descriptor.go
```diff
@@ -0,0 +1,53 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("Invalid resource descriptor: %w", err)
+	}
+
+	return rd, nil
+}
+
+/*
+RDListFromRecord converts a map of artifacts as collected
+by RecordArtifacts() in runlib.go and converts it into a list
```
adityasaky, line 36: Can this be expanded with the structure of the map? We may retire RecordArtifacts at some point, it'll be helpful to track this here.

in_toto/attestation/v1/resource_descriptor.go
```diff
@@ -0,0 +1,53 @@
...
+		Digest:           digestSet,
+		Content:          content,
+		DownloadLocation: downloadLocation,
+		MediaType:        mediaType,
+		Annotations:      annotations,
+	}
+
+	err := rd.Validate()
+	if err != nil {
+		return nil, fmt.Errorf("Invalid resource descriptor: %w", err)
```
adityasaky, line 28:
```suggestion
		return nil, fmt.Errorf("invalid resource descriptor: %w", err)
```

I think we need to enable a linter.

in_toto/attestation/v1/statement.go
```diff
@@ -0,0 +1,29 @@
...
+	st := &ita1.Statement{
+		Type:          ita1.StatementTypeUri,
+		Subject:       subject,
+		PredicateType: predicateType,
+		Predicate:     predicate,
+	}
+
+	err := st.Validate()
+	if err != nil {
+		return nil, fmt.Errorf("Invalid in-toto Statement: %w", err)
```
adityasaky, line 25:
```suggestion
		return nil, fmt.Errorf("invalid in-toto Statement: %w", err)
```

in_toto/runlib.go
```diff
@@ -339,22 +339,37 @@ func InTotoRun(name string, runDir string, materialPaths []string, productPaths
...
+			Name:        name,
+			Materials:   materials,
+			Products:    products,
+			ByProducts:  byProducts,
+			Command:     cmdArgs,
+			Environment: map[string]interface{}{},
+		}
 	}
 
 	if useDSSE {
```
adityasaky, line 368: FWIW, this should apply if `useDSSE` or `attest` is set. We don't want `attest` to go to the legacy sig wrapper.

adityasaky (changes requested):
Some nits. Should we discuss this in the context of #288 as well? cc @Forrin

in_toto/runlib.go
marcelamelara: Ok! I wasn't sure if this was the intent here. I'll change this.

in_toto/attestation/predicates/provenance/v1/provenance.go
```diff
@@ -0,0 +1,98 @@
...
+function returns an error.
+*/
+func GenerateValidPredicate(buildDefinition *prov1.BuildDefinition, runDetails *prov1.RunDetails) (*prov1.Provenance, error) {
+	provenance := &prov1.Provenance{
+		BuildDefinition: buildDefinition,
+		RunDetails:      runDetails,
+	}
+
+	if err := provenance.Validate(); err != nil {
+		return nil, fmt.Errorf("Invalid Provenance format: %w", err)
```
marcelamelara, line 94:
```suggestion
		return nil, fmt.Errorf("invalid Provenance format: %w", err)
```

in_toto/runlib.go
marcelamelara, line 368:
```suggestion
	if useDSSE || attest {
```

marcelamelara:
```suggestion
	if useDSSE || attest {
```

in_toto/attestation/v1/resource_descriptor.go
marcelamelara: Just to be 100% sure I've understood what you're asking, you're suggesting we include `map[string]map[string][string]` in this comment?

in_toto/attestation/common/common.go
marcelamelara: I think I had set it public since it's called from within `runlib`, but either moving this to `internal/` or even as a private function to be called only within `GenerateValidStatement()` could work. We would have to change the signature of `GenerateValidStatement()` to take a protobuf message instead of Struct in the latter case, not sure how I feel about that.

Forrin:
> Some nits. Should we discuss this in the context of #288 as well? cc @Forrin

There is for sure some overlap. These can probably be merged fairly easily if I make a few adjustments in my PR. I can remove calling the Attestor and leave that for a future change. Whether it's done in the run function or somewhere else I don't have a strong opinion. However, I do think that the run function is getting a bit bloated and doing too much. Possibly passing it a list of attestors would make the most sense.

But, the run function is technically for a Link style metadata... which I view as an attestor. Open to further conversations.

In the short term, removing how the Attestor code is called and leaving only the interface and LinkAttestor implementation would at least get it merged.

@adityasaky @marcelamelara

marcelamelara:
This has been superseded by https://github.com/in-toto/go-witness/pull/149 .

PR #333, opened 2024-06-03
Resolve all the error in unit tests

DarikshaAnsari:
**Fixes issue:#322**

**Description:**

requirements:**


lukpueh:
Ci fails. can you check? 

FYI: you can also run CI when you push to your fork, I think you'll need to remove the branch restriction in the workflow file for that:
https://github.com/in-toto/in-toto-golang/blob/72262c0fd3a9c392440040c04336174eddac427b/.github/workflows/build.yml#L1-L3

PR #347, opened 2024-09-05
Update coveralls to v2

shibumi:
Fixes broken coveralls Github action pipeline step.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -20,7 +20,7 @@ jobs:
     - name: Test
       run: go test -covermode atomic -coverprofile='${{ matrix.os }}-${{ matrix.go-version }}.cov' ./...
     - name: Coveralls Parallel
-      uses: coverallsapp/github-action@643bc377ffa44ace6394b2b5d0d3950076de9f63
+      uses: coverallsapp/github-action@v2
       with:
         file: '${{ matrix.os }}-${{ matrix.go-version }}.cov'
         parallel: true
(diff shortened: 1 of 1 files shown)


shibumi:
@lukpueh  @adityasaky I have modified the coveralls setting a bit to accommodate for development.
We now accept everything above 70% and changes can have a drift of -1%.

I doubt we will move the project back to 90% again in near future. The bad value comes from the untested command line tool.

AMRALAA10 (approved):

.github/workflows/build.yml
```diff
@@ -20,7 +20,7 @@ jobs:
     - name: Test
       run: go test -covermode atomic -coverprofile='${{ matrix.os }}-${{ matrix.go-version }}.cov' ./...
     - name: Coveralls Parallel
-      uses: coverallsapp/github-action@643bc377ffa44ace6394b2b5d0d3950076de9f63
+      uses: coverallsapp/github-action@v2
```
adityasaky, line 23: We should probably try and pin to hash, dependabot supports those bumps.

Also, 643bc377ffa44ace6394b2b5d0d3950076de9f63 is the latest release commit AFIACT? https://github.com/coverallsapp/github-action/releases/tag/v2.3.0

So, do we need this diff?

adityasaky (commented):
Thanks for fixing this @shibumi! Was the threshold change on the coveralls dashboard? If so, we should also work out a way for a few of us to have access maybe?

.github/workflows/build.yml
shibumi: Mh, I missed this. In this case we do not need this PR. The fix on the dashboard fixed the coverall problems

shibumi:
@adityasaky  I think you have already access. You just have to login to the coveralls UI via your GitHub account. If not, please ping me and I can take a look at the settings

shibumi:
Closing this, because coveralls is already up to date.

Issue #369, opened 2024-12-06
Deprication of in-toto-golang - Tracker


06kellyjac:
The plan for a while now has been to deprecate https://github.com/in-toto/in-toto-golang in favor of the https://github.com/in-toto/go-witness library.

Actions required for deprecation and eventual archiving of the repo:

* [X] Create a tracker issue for recording progress/status
* [ ] Pin the tracker issue
* [ ] Add identifiers to the code to be found by code quality scanners
* [ ] Add a notice on the README
* [ ] Add simple examples to facilitate migrations
* [ ] Discuss cutting a v1 of in-toto/go-witness
* [ ] Decide a date to stop updating dependencies
* [ ] Decide a date to archive the repo (locks issues etc)
* [ ] Archive

Light discussion is welcome in the issue. If conversations are very active we will look to break out into separate descriptions

shibumi:
I pinned the tracker issue.

PR #401, opened 2025-03-29
Deprecate all packages in in-toto-golang

shibumi:
This commit deprecates all packages within in-toto-golang and the most important methods.


in_toto/keylib.go
```diff
@@ -285,6 +285,8 @@ On success it will return nil. The following errors can happen:
   - no valid PKCS8/PKCS1 private key or PKIX public key
   - errors while marshalling
   - unsupported key types
+
+Deprecated: This method has been deprecated.
```
06kellyjac, line 289: I understand having just "This method has been deprecated." for where there's a whole package deprecation note.
But for packages where only specific functions are deprecated so far do we want a longer comment which links to go-witness?

06kellyjac (commented):
progress on #369 

Glad to see this moving, just one thought.

in_toto/keylib.go
shibumi: I am not sure how far we want to go. At first I thought of putting the deprecation notice only on package level, but then I realized that many IDEs and code editors do not pick up these deprecation notices and they only pick it up for method comments.

Adding a comment to EVERY method seems really toilsome for me, so I tried adding it only to the most important method to make clear that we deprecate this.

We can link to go-witness from there as well, but I kinda want to avoid putting a big notice on every exported variable and method.

adityasaky:
Maybe a slightly reckless solution, but I'm half tempted to cut a /v2 module release with just the stuff removed to see where the complaints are. My sense is that most of the consumption of this repo is for provenance / attestation structs and I'm curious if the workflows are in fact used. If we cut a breaking release and bump the major version, we can at least see who gets broken and determine if we want to cut subsequent 1.x releases?

PR #462, opened 2026-05-04
match: Replace ^ with ! for negation in character classes

adityasaky:
This matches the behavior of fnmatch in in-toto/in-toto.

**Fixes issue:**

N/A

**Description:**

There is an inconsistency with in-toto-golang's glob matching compared to in-toto-python. The negation character used within character classes in in-toto-python is `!` and here it was `^`. This PR makes it `!` to be consistent with in-toto-python.

requirements:**


jkjell (approved):

in_toto/match.go
```diff
@@ -22,8 +22,11 @@ var errBadPattern = errors.New("syntax error in pattern")
 //	term:
 //		'*'         matches any sequence of non-/ characters
 //		'?'         matches any single non-/ character
-//		'[' [ '^' ] { character-range } ']'
+//		'[' [ '!' ] { character-range } ']'
 //		            character class (must be non-empty)
+//
+// NOTE: Only '!' is supported for character class negation, not '^'
```
JustinCappos: Maybe also explain why: to make this compatible across language ecosystems

JustinCappos (approved):
looks good.  Minor comment suggestion

in_toto/match.go
adityasaky: done

Issue #466, opened 2026-06-13
VerifyLinkSignatureThesholds non-deterministically rejects valid links when a step mixes pubkey and certificate functionaries


kanywst:
**General information:**

* Version (tag or commit): `36d782ffb2ca3adbffcdce1fd971c23319dd4469`
* Operating system (Linux/Mac/Windows): Mac
* Go version: go1.26.4
* Go build flags (if any): none

**Description of the bug:**

For a step whose functionaries are a mix of `pubkeys` and certificate constraints, `VerifyLinkSignatureThesholds` returns success or a "not enough links" error non-deterministically for the same valid input.

The cause is that `isAuthorizedSignature` is declared once before the per-link loop instead of inside it (`in_toto/verifylib.go:563`), so after a pubkey link sets it true, later certificate-functionary links skip verification. Whether a certificate link counts toward the threshold then depends on whether it is iterated before or after a pubkey link.

Reproduction: build one link signed by a pubkey functionary and one signed by a certificate functionary, put both in a step with `Threshold: 2`, and run `VerifyLinkSignatureThesholds` repeatedly. Result over 50 runs of the same input:

```text
over 50 runs of the SAME valid input: pass(threshold met)=4 fail(false-reject)=46
```

A self-contained test is in the PR.

This is not a verification bypass (an unauthorized link is never accepted). The effect is that a valid supply chain intermittently fails verification.

**Anything special you want to tell us?**

Fix is one line: move `isAuthorizedSignature := false` inside the per-link loop so it resets each iteration. PR attached.

webdevsamran:
Hi team, I would like to work on this verification threshold issue.

In `VerifyLinkSignatureThesholds`, mixing public key and certificate functionaries causes non-deterministic verification failures due to inconsistent key-id map lookups across functionary types. I can update the functionary signature matching loop in `in-toto/in_toto.go` to deterministically verify signatures against both key formats, with unit tests.

Could you please assign this to me? Thanks!

PR #467, opened 2026-06-13
Reset isAuthorizedSignature per link in threshold verification

kanywst:
**Fixes issue:** #466

**Description:**

`isAuthorizedSignature` was declared once before the per-link loop instead of inside it, so after a pubkey link set it true, later certificate-functionary links skipped verification. For a step mixing pubkey and certificate functionaries the threshold result then depended on map iteration order (4 passes, 46 false-rejects over 50 runs of the same valid input).

This moves the declaration inside the loop. The added test builds such a step at threshold 2 and verifies it 20 times; it fails before the change and passes after.

requirements:**


kunalworldwide (commented):
Good catch on a real correctness bug. Moving `isAuthorizedSignature = false` inside the inner loop means each link gets evaluated independently — without this fix, once a pubkey-based link sets the flag to `true`, subsequent certificate-based links skip their cert verification entirely because the flag is still `true` from the previous iteration.

The test is well-constructed: it creates a mixed-functionary step with both a pubkey signer and a certificate signer, then verifies that the threshold check works correctly regardless of map iteration order. The `for range 50` loop is a pragmatic way to exercise the non-deterministic ordering.

One observation: the variable `isAuthorizedSignature` was declared outside the outer loop (before line 563 in the original), so in theory the bug could also have leaked state across *steps*, not just across links within a step. By moving the declaration inside the `for signerKeyID` loop, you've fixed both cases.

LGTM.

PR #471, opened 2026-07-26
verifylib: don't abort verification on a single failed sublayout

### LINKED ISSUE #23: Failure of sublayout verification leads to failure of entire verification routine
*Please fill in the fields below to submit an issue or feature request.  The
more information that is provided, the better.*

#19 introduces sublayout verification capabilities, which closely resembles the reference implementation. Currently, the overall verification process fails as soon as sublayout verification fails at some stage, instead of proceeding with other steps and checking for threshold value later.

**Expected behavior**:

The implementation must continue the verification routine and consider summary links only from successfully verified sublayouts. The verification routine can then check if threshold value is met further down the line.

For more details: #19#discussion_r255517008

pujitha24:
Motivation:

VerifySublayouts recursively verifies each functionary's sublayout
link for a step, and previously returned an error as soon as any one
of those recursive verifications failed. This aborted the entire
top-level InTotoVerify call even when other functionaries for that
same step already had valid links that would satisfy the step's
signature Threshold. This is inconsistent with
VerifyLinkSignatureThesholds, which already tolerates individual bad
signatures as long as enough of the others meet the threshold.

Approach:

When a functionary's sublayout verification fails, drop that
functionary's link instead of returning immediately, and continue
verifying the rest. After all functionaries for a step are processed,
check the step's Threshold against however many links (regular or
resolved sublayout summary links) actually survived; only then return
an error if the step no longer has enough.

Impact: user-visible behavior is unchanged for the common case where
all functionaries succeed, or where a single functionary's failure
already caused the step to fall below Threshold (still an error,
though the message now reports the shortfall after resolving
sublayouts, rather than the first sublayout error verbatim). The
concrete behavior change is: a step with redundant sublayout
functionaries (Threshold reachable without all of them) now verifies
successfully instead of failing outright when only one of them is
bad.

Validation:

go build ./... && go test ./... - all packages pass, including the
existing TestVerifySublayouts and the new
TestVerifySublayoutsPartialFailure, which injects a bogus second
functionary link for a sublayout step and asserts that verification
still succeeds on the surviving, valid link while the failed one is
dropped.

Fixes #23

Signed-off-by: Pujitha Paladugu <10557236+pujitha24@users.noreply.github.com>


pujitha24:
Just checking in on this one — it's still green and rebased, happy to make any changes that would help it move through review.

pujitha24:
Closing this to keep my open PR queue manageable - it has been open a while without review, and I would rather not leave stale PRs sitting in your queue. The change itself still applies; happy to reopen and rebase if it is useful to you.

